PR #4, opened 2015-05-09
Add ImageView widget + add ImagePanel::images() accessor

mrzv:
This request adds a simple ImageView widget and modifies the example to show an image using it (and update on selection from the image panel).


wjakob:
this looks useful, thank you. I have a few minor improvements to ensure that things are shown with the correct aspect ratio and when the image is not initialized (see the next commit).

PR #5, opened 2015-05-14
Add ColorWheel widget for color picking

mrzv:
The drawing commands are taken directly from the NanoVG demo.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,6 +108,7 @@ add_library(nanogui STATIC
   include/nanogui/imagepanel.h src/imagepanel.cpp
   include/nanogui/imageview.h src/imageview.cpp
   include/nanogui/vscrollpanel.h src/vscrollpanel.cpp
+  include/nanogui/colorwheel.h src/colorwheel.cpp
   ${extra_source}
 )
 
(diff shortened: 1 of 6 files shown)


wjakob:
Hi,

that looks cool, though there are still some things I'd like to see fixed before merging.

Right now, the example application looks really ugly after the change :) -- partly because the color wheel is so huge, and it messes up the layout for everything else.

When fixing the wheel to a smaller size (e.g. 100x100), the proportions don't seem right anymore.

Also, it would be best if the callback takes the updated color value as an argument like in some other widgets (e.g. Slider).

Thanks,
Wenzel

mrzv:
I'll play around with the sizes. To be honest, currently the relationship between preferred layout, fixed layout, and the actual layout are a little mysterious to me.

I originally gave the color value to the callback, but then I changed it to the current way. The reason is that I didn't like spelling out `Vector3f`, and it's quite easy to capture the color wheel in the lambda and just call `color()`.

BTW, I'd love to be have a color picker button (that's colored the way of the chosen color) that when pressed opens up a color wheel. Currently, I just open up a Popup and put the color wheel into it. But if you have the energy to make it so that one could open the color wheel by itself (without a popup underneath), that would  be nice. Currently, it's outside my comfort level with nanogui.

wjakob:
Awesome, thanks. I made a few more minor changes related to formatting and event handling (in the next commit).

PR #6, opened 2015-06-23
Pull request 

schuellc:
new FormScreen layout class
editable TextBox
added new GridLayout
mouse cursor and fontSize can be set per widget
switched to strongly typed enums
various features and bugfixes


wjakob:
Looks like this pull request is hard-coded to go to nanogui:master, and it's impossible to change the target branch. I manually merged into a new branch named libigl-updates and will close the pull request as it is no longer needed.

PR #17, opened 2015-10-13
screen resize callback

nickolasrossi:
I had a need to get the glfw window resize event passed down into my Screen. Seemed straightforward to add and hopefully useful.


wjakob:
looks good, thanks.

wjakob:
FYI: I changed the API a bit and changed the callback so that it also reacts to framebuffer size changes. this was needed to support moving windows between retina and non-retina displays on OSX.

PR #19, opened 2015-10-23
Add ImageView::SizePolicy to optionally scale images up to the ImageView size (fixes #18)

### LINKED ISSUE #18: Allow images to scale up to the size of ImageView
Currently, if the size of `ImageView` is larger than the image, the image will be shown at its native (smaller) size. It would be nice to (maybe, optionally) remove this restriction. The actual change to the code of `ImageView::draw` requires removing (or rather adjusting a little) the checks for `s.x() < w` or `s.y() < h`.

mrzv:


wjakob:
Looks good, but two more requests: can you change the constructor so that it has a default value matching the previous behavior? (to avoid breaking code).

Secondly, python bindings are missing for the two new functions (needs two more lines in python/python.cpp)

Thanks,
Wenzel

mrzv:
I'll add the python bindings shortly, but I'm confused by the constructor comment. The previous behavior was equivalent to the `Fixed` policy (i.e., don't grow past full image size). That's the constructor default now.

wjakob:
never mind -- I probably looked at the constructor implementation by accident, which doesn't have the default parameter.

I just added one more commit which changes the scoping so that the python API resembles the C++ enum class more closely.

PR #21, opened 2015-11-14
A few fixes

r-lyeh:
- Fix _WIN32 directive 
- Do not redefine GLEW_STATIC if already present (ie, when embedding NanoGUI in an external project)
- Fix MSVC snprintf (defaults to _snprintf)
- Fix crash if no slashes are present at given filenames (bin2c.c)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,6 +49,11 @@ endif()
 # Compile GLFW
 add_subdirectory("${CMAKE_CURRENT_SOURCE_DIR}/ext/glfw" "ext_build/glfw")
 
+# Python support: add NANOGUI_PYTHON flag to all targets
+if (NANOGUI_BUILD_PYTHON)
+  add_definitions ("-DNANOGUI_PYTHON")
+endif()
+
 # Shared library mode: add NANOGUI_SHARED flag to all targets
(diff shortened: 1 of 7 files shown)


r-lyeh:
Btw, suggestion:

`python.cpp` should have a configuration guard so the file is compiled only if pybind11 support is enabled. Something like:

``` 
#ifdef NANOGUI_USE_PYTHON_BINDINGS
// original python.cpp here
#endif
(EOF)
```

wjakob:
Looks all good except for a minor thing (see my comment).

I didn't quite understand what you said regarding the python include guard. If python support is disabled, the file will be completely dropped by the cmake build system, no? (i.e. it won't even be in the project)

r-lyeh:
1) Yep, will do asap
2) Yes theorically, unless you are using another build system (like premake) or embedding the sources into your project (as I did).

wjakob:
ok -- I'd be happy to accept a patch for 2) as long as it works for the cmake build as well.

r-lyeh:
I think I'm done. 
I've tried to replicate your style for the directives in the `CMakeLists.txt` file as much as possible.

wjakob:
Looks good, I committed it -- thanks!

r-lyeh:
:+1:

PR #23, opened 2015-11-18
made Screen independent of glfw by splitting into ScreenCore & Screen

giorgiomarcias:
I needed nanogui without GLFW management as I already have event handlers etc. (see my project GLFWM),  even if Screen already provided callbacks to register for. So I split it into ScreenCore (managing widgets) and Screen (managing GLFW callbacks).
There is also a new time setting for tooltip duration.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,12 +2,6 @@ cmake_minimum_required (VERSION 2.8.3)
 
 project("NanoGUI")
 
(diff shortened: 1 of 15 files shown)


wjakob:
Not sure I am comfortable integrating such a big redesign for a 1-off requirement :-S. Maybe it is better if you publish a separate fork?

giorgiomarcias:
Ok. Thanks anyway

PR #39, opened 2015-12-24
added Widget::add function with variadic semantic

dalerank:
Widget::add hide "new operator" in source code, and able to construct form more simple

### DIFF
--- a/include/nanogui/widget.h
+++ b/include/nanogui/widget.h
@@ -27,6 +27,13 @@ enum class Cursor;
  */
 class NANOGUI_EXPORT Widget : public Object {
 public:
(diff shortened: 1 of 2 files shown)


wjakob:
This is a neat use of C++11. The reasons for not merging it are that
1. it doesn't really reduce the amount of code that has to be written
2. it obscures the piece of code that is actually invoked (i.e. a constructor)

wjakob:
Hi dalerank,

I had a change of mind :)  --- due to another upcoming widget (TabWidget) where this sort of functionality is useful, I added a modified version of your variadic construction syntax.

Thanks,
Wenzel

PR #41, opened 2015-12-30
Correct Python shutdown definition

mrzv:


wjakob:
Oops, thanks for catching that!

Issue #48, opened 2016-03-03
mTheme accessed in some widget's constructor


romp13:
Hi Wenzel,

Some widgets (in particular Label and Textbox) access mTheme in their constructor. mTheme is however copied from the parent widget otherwise it is a null pointer. This means you cannot create such a widget without parent and attach it later to the GUI. This not a bug per say but it would be convenient as I am doing this regularly in my code.

Thanks!

wjakob:
Hi Romain,

good point, I see how this could be an issue. Do you have any suggestions on how to fix it?

Best,
Wenzel

romp13:
I guess the less invasive fix would be to just check if mTheme is not null in these constructors before accessing these properties and if I want to attach it later I need to set these properties manually.
The better version of this would be to override setParent and/or setTheme for the affected widgets to copy these properties upon attachment to a parent. And I guess this should propagate to children also because otherwise compound widget could be inconsistent. What do you think?
Do you have another idea?
I could prepare a pull request if needed.

Thank you!

wjakob:
Not checking for mTheme seems hacky -- it's also likely that more kinds of properties will migrate from hardcoded constants into the theme in the future.

Having a ``setTheme`` function sounds like a good plan to me (which recursively calls ``setTheme`` on the child widgets). ``setParent`` is less suitable because the parent would just change in one place, so it' odd to propagate that call to the child widgets as well.

romp13:
I agree this sounds like the best option. 
Let me know if I should prepare a pull request or if you prefer to take care of it yourself.

wjakob:
It's not a priority for me. If it causes issues for you then please go ahead and make a PR.

For this kind of interface change, my request would be that you do the change consistently for all widgets :).

wjakob:
also, watch out for the coding style -- no tabs, braces on the same line, etc. Often PRs have to be redone because they are inconsistent with the remainder of the codebase.

romp13:
Yes I'll try my best.
Quick question: should setTheme override properties that have been set
manually? For example for Label should setTheme override mFontSize in all
cases or only if mFontSize < 0? If you prefer the second alternative how to
handle the mColor? Should I checked for the alpha channel being strictly
positive to decide if it has been set manually or not?

wjakob:
PR is finally merged... :)

PR #49, opened 2016-03-03
mTheme accessed in some widget's constructor -- issue #48

romp13:
Unfortunately the `setTheme` method being virtual cannot be used in the Widget constructor, that is why you need to keep setting mFontSize and so on both in constructor and setTheme. The other option would be to remove setThemefron the Widget constructor and call it in every widget's constructors, but in my opinion this is less appropriate.

### DIFF
--- a/include/nanogui/label.h
+++ b/include/nanogui/label.h
@@ -41,6 +41,9 @@ public:
     /// Set the label color
     void setColor(const Color& color) { mColor = color; }
 
+    /// Set the \ref Theme used to draw this widget
+    virtual void setTheme(Theme *theme) override;
+
     /// Compute the size needed to fully display the label
     virtual Vector2i preferredSize(NVGcontext *ctx) const;
     /// Draw the label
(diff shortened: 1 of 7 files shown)


wjakob:
What about calling setTheme from addChild()?

romp13:
That avoids me to call addChild and then setTheme to attach the widget,
however this doesn't solve the virtual method problem because the widget
constructor is calling addChild who is then calling Widget::setTheme, not
the override version.
Should I override my previous PR or just add a new commit? (I don't know if
you can merge the PR into a single commit)

wjakob:
You can either ``commit --amend`` it and then force-push, or I will squash everything into one commit on my end.

romp13:
Somehow amending failed and it created another commit, maybe because it's already referenced by github as a pull request. So you'll need to squash it on your end, sorry for that.
Let me know if you are satisfied with this version or if you would do something differently.
Thank you for your help in fixing this.

romp13:
Hey Wenzel,
Just to check why you didn't merge the pull request yet. Is there a problem with my changes?
Thanks!

wjakob:
Hi Romain,

apologies for the huge delay in getting this accepted. I basically had to go through all of my projects and ensure that your change doesn't break anything there, and it's been tricky to find time for that.

Thanks,
Wenzel

PR #51, opened 2016-03-07
Application main loop without the 50ms timer

tewk:


wjakob:
Hi,

I'd prefer if the mainloop function took a parameter with a default value on the number of milliseconds. Then that could be e.g. set to "-1" to indicate that no regular redraw events should be sent. In that case, plase also add a line or so to the comment which explains what the modified function does.

Wenzel

PR #59, opened 2016-05-03
Initialize mBackground before first glClearColor call in constructor,…

sjm324:
… remove now redundant call from Screen::initialize.

I was debugging my own program and `valgrind` gives a bunch of `Conditional jump or move depends on uninitialised value(s)` because `glClearColor` is called with `mBackground` before it is initialized.

Issue can be replicated with `example1` in previous commit on master.  I applied it to both constructors, but I'm not entirely sure what the empty constructor is used for so it may not be appropriate.  The `glClearColor` call on line ~130 of `screen.cpp` may not be necessary since `drawAll()` will use it too.

An alternative would be to keep `mBackground` getting set in `initialize()` and just not call clear there?  I'm not sure which is more fitting to the framework / how all of the pieces fit together, but I hope this helps :)

### DIFF
--- a/src/screen.cpp
+++ b/src/screen.cpp
@@ -72,15 +72,16 @@ static float get_pixel_ratio(GLFWwindow *window) {
 }
 Screen::Screen()
     : Widget(nullptr), mGLFWWindow(nullptr), mNVGContext(nullptr),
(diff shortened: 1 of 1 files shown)


wjakob:
Good catch -- thanks!

PR #65, opened 2016-05-12
Enabling support for alternative core profiles i.e. 4.1. User is resp…

sjm324:
…onsible for requesting a forward compatible core profile if changing the default 3.3 core.

I ended up needing certain features for `4.x` and related `ARB` extensions in another project.  I don't really understand the full implications of this form of update so it may not be stable / appropriate to include in the upstream.

1. Default profile requested is still core 3.3.
2. Screen default arguments in constructor can be changed to request e.g. 4.1.
3. Requests for forward compatible and core profile remain as is (and will fail if an invalid profile is requested).
4. A not so clean example of how to use it in `example2`.  There is almost certainly a way to do this using more elegant C++11 features, but I don't know how to only specify two default args at the end.

### DIFF
--- a/include/nanogui/screen.h
+++ b/include/nanogui/screen.h
@@ -61,7 +61,7 @@ public:
      * \param glMajor
      *    The requested OpenGL Major version number.  Default is 3, if changed
      *    the value must correspond to a forward compatible core profile (for
-     *    portability reasons).  For example, set this to 4 and glMinor to 1
+     *    portability reasons).  For example, set this to 4 and \ref glMinor to 1
      *    for a forward compatible core OpenGL 4.1 profile.  Requesting an
      *    invalid profile will result in no context (and therefore no GUI)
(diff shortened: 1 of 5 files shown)


wjakob:
Can you also add these extra arguments and their default values to the Python bindings? Otherwise, you'll break the included example projects.

sjm324:
Oops sorry!  Just added them, `example2` was mirrored to match the cpp side.  It works fine for me on linux and osx, but I am having trouble building it on Win64 with MSVC 2015; error [C2228](https://msdn.microsoft.com/en-us/library/3y365xw6.aspx). I don't think I caused that, but it may be worth verifying.  I'm also not the greatest at getting things to build on Windows...

wjakob:
Looks great, thank you. The CI build bots didn't complain, so I suppose the C2228 error was something on your end.

PR #66, opened 2016-05-17
Suppress warnings from external libraries

jslee02:
[Compiling Eigen with modern compilers generates a lot of warnings as `class std::binder2nd` is deprecated](http://stackoverflow.com/questions/30535933/eigen-gcc-5-class-stdbinder2nd-is-deprecated), and [it seems it's fixed in upstream](http://eigen.tuxfamily.org/bz/show_bug.cgi?id=872). So the warnings could be suppressed if `nanogui` use the patched Eigen version. However, I would suggest skipping warnings from external libraries in general because it's their responsibility.

This PR switch `nanogui` to skip the warnings from external libraries using [CMake's `SYSTEM` option in `include_directories`](https://cmake.org/cmake/help/v3.0/command/include_directories.html).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,7 +122,7 @@ if (NANOGUI_USE_GLAD)
     set_source_files_properties("${CMAKE_CURRENT_SOURCE_DIR}/ext/glad/src/glad.c"
       PROPERTIES COMPILE_FLAGS "/wd4055 ")
   endif()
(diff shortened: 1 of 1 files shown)


wjakob:
That seems like a rather hacky solution IMHO -- these are not system header files. The better way will be to wait for Eigen 3.3, which is out any day now.

PR #68, opened 2016-05-17
Enable coroutine support on every platform

mrzv:


wjakob:
A comment: let's enable coroutines on Mac and Linux, but not on Windows.

wjakob:
Processing events on a different thread there is fine AFAIK.

wjakob:
(I think I actually tested your example3 at some point)

mrzv:
How do you check for Linux in your code?

wjakob:
```
#if defined(__linux__)
...
#endif
```

mrzv:
Ok, I think I got it now.

wjakob:
thanks!

PR #69, opened 2016-06-08
quick edit for IntBox, FloatBox and ComboBox

romp13:
Hi Wenzel,

I added some simple features to more quickly edit the value of IntBox, FloatBox and ComboBox using scrolling and dragging which makes it much more convenient than typing a value. Here are the full list:

- added scrollEvent support for ComboBox
- added right dragging control for IntBox and FloatBox (à la Blender)
- added scrollEvent for same effect
- added double right click to reset to default value for TextBox

Let me know if you have some feedback for improvements. You can easily test these features on Example 1 or 2.
Also I did not add the Python bindings for these. When you approve the changes, you or I can add them quickly to the PR.

Thanks,
Romain

### DIFF
--- a/src/screen.cpp
+++ b/src/screen.cpp
@@ -485,7 +485,7 @@ bool Screen::mouseButtonCallbackEvent(int button, int action, int modifiers) {
             glfwSetCursor(mGLFWWindow, mCursors[(int) mCursor]);
         }
 
-        if (action == GLFW_PRESS && button == GLFW_MOUSE_BUTTON_1) {
+        if (action == GLFW_PRESS && (button == GLFW_MOUSE_BUTTON_1 || button == GLFW_MOUSE_BUTTON_2)) {
             mDragWidget = findWidget(mMousePos);
             if (mDragWidget == this)
                 mDragWidget = nullptr;
--- a/src/textbox.cpp
+++ b/src/textbox.cpp
@@ -17,6 +17,7 @@
 #include <nanogui/textbox.h>
 #include <nanogui/opengl.h>
 #include <nanogui/theme.h>
+#include <nanogui/entypo.h>
 #include <nanogui/serializer/core.h>
 #include <regex>
 
@@ -25,6 +26,7 @@ NAMESPACE_BEGIN(nanogui)
 TextBox::TextBox(Widget *parent,const std::string &value)
     : Widget(parent),
       mEditable(false),
+      mSpinnable(false),
       mCommitted(true),
       mValue(value),
       mDefaultValue(""),
@@ -68,9 +70,13 @@ Vector2i TextBox::preferredSize(NVGcontext *ctx) const {
     } else if (!mUnits.empty()) {
         uw = nvgTextBounds(ctx, 0, 0, mUnits.c_str(), nullptr, nullptr);
     }
+    float sw = 0;
+    if (mSpinnable) {
+        sw = 14.f;
+    }
 
     float ts = nvgTextBounds(ctx, 0, 0, mValue.c_str(), nullptr, nullptr);
-    size(0) = size(1) + ts + uw;
+    size(0) = size(1) + ts + uw + sw;
     return size;
 }
 
@@ -91,8 +97,10 @@ void TextBox::draw(NVGcontext* ctx) {
     nvgRoundedRect(ctx, mPos.x() + 1, mPos.y() + 1 + 1.0f, mSize.x() - 2,
                    mSize.y() - 2, 3);
 
-    if(mEditable && focused())
+    if (mEditable && focused())
         mValidFormat ? nvgFillPaint(ctx, fg1) : nvgFillPaint(ctx, fg2);
+    else if (mSpinnable && mMouseDownPos.x() != -1)
+        nvgFillPaint(ctx, fg1);
     else
         nvgFillPaint(ctx, bg);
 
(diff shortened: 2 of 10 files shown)


wjakob:
Hi Romain,

for Int and FloatBox, I find this sort of interaction quite unexpected. I think that it would be better to factor it out into a different widget that indicates the ability to make tweaks -- this is usually referred as a Spinner or SpinBox (http://manual.qooxdoo.org/1.3/_images/spinner.png). It would be possible fairly easily I think by subclassing TestBox.

For the ComboBox, I was surprised that the selection doesn't go down when rolling the mouse wheel down -- maybe a flipped sign, or is this the usual "natural" vs "non-natural scroll direction" discussion?

Best,
Wenzel

romp13:
I personally find it very convenient because when you expose a value to the GUI you either want to be able to give it a specific value (which is what you already provided) and/or want to vary this value continuously and see its effect (which currently you cannot easily and is what this functionality brings you). In my opinion, specializing into another widget adds to the complexity of the library, I would rather make it an option for IntBox and FloatBox than you can enable/disable. Also because of the FormHelper: how do you know when you call addVariable<int>(...) whether you want a IntBox or a SpinBox?
What do you think of having it optional? (you could even draw arrows on the side to visually see the difference)

For the scrolling direction, this is usually a preference you can set in your system but I don't know if this is taken into account by GLFW and reflected in the sign of rel.y(). If this is the case we can choose which corresponds to which, and if not we could have an option for the screen. Alternatively you could just choose which one you prefer and we stick to this.

Best,
Romain

wjakob:
Hi Romain,

agreed -- we could leave it in Int/FloatBox, but having clickable arrow buttons would be nice if this option is active.

For the scroll direction, it would make sense to stick with what the library has already (which is the other way I think) -- you could e.g. compare to the ScrollWidget.

Thanks,
Wenzel

romp13:
Sounds like a plan :-)
I'll push a modified version as soon as possible.

Thanks,
Romain

src/screen.cpp
```diff
@@ -485,7 +485,7 @@ bool Screen::mouseButtonCallbackEvent(int button, int action, int modifiers) {
             glfwSetCursor(mGLFWWindow, mCursors[(int) mCursor]);
         }
 
-        if (action == GLFW_PRESS && button == GLFW_MOUSE_BUTTON_1) {
+        if (action == GLFW_PRESS && (button == GLFW_MOUSE_BUTTON_1 || button == GLFW_MOUSE_BUTTON_2)) {
```
wjakob: One more thing -- is this related to the rest of the commit, or did it slip in by mistake?

src/textbox.cpp
```diff
@@ -250,7 +250,28 @@ bool TextBox::mouseButtonEvent(const Vector2i &p, int button, bool down,
                 /* Double-click: select all text */
                 mSelectionPos = 0;
                 mCursorPos = (int) mValueTemp.size();
-                mMouseDownPos = Vector2i(-1, 1);
```
wjakob: Is the sign change in the Y axis intentional here?

romp13: For the first question, I found it necessary to add this otherwise I could
not drag with the right click. I did not want to use the left click as it
would interfere with the default behavior. But this is actually good
because currently the dragging can only be achieved with left click.

For the second thing, I think it was a typo in your code because I could
not find the reason for this. But I think it doesn't impact anything. Or
was it intentional? It is basically to reset it mouseDown right? Everywhere
else you use -1,-1.

wjakob:
FYI: Kenshi has added some kind of spin box in his nanogui fork -- so that may be worth taking a look at. :)

wjakob:
cc @kenshi84

romp13:
I finally found a little bit of time to finish this. Here are the changes I made:
- ComboBox scrolling is now in the correct direction
- Spinning is now optional and if enabled arrows appear on the left side
- Cursor switch to hand when hovering the arrows
- Spinning is controlled by clicking one arrow for single step, or mouse wheel or right click drag for multi-step
- IntBox and FloatBox have now a min and max options (by default -inf and +inf) which are convenient when scrolling

Let me know if you are happy with this version.

wjakob:
Did you take a look at Kenshi's version also? It would be interesting to understand the tradeoffs.

romp13:
Yes I checked Kenshi's version. He created NumberSlider and SpinBox which combine an IntBox or FloatBox respectively with a slider or two buttons (+ and -). The big drawback of this solution is that it creates multiple widgets with different control capabilities whereas my solution brings all these controls into a single widget (as an option). In particular his FormHelper also provides addSlider and addSpinBox in addition to addVariable which I find quite inconvenient. If you expose a scalar to your GUI you expect 3 controls: direct editing of the value (which is what the IntBox and FloatBox are already providing), discrete stepping (which is what Kenshi's SpinBox does, or my improved TextBox in a similar way), continuous sliding (which is what Kenshi's NumberSlider offers with an extra slider (but only possible if min and max values are specified), or my improved TextBox via mouse scrolling and mouse dragging).
The only drawback I see for my solution is that some users might not be familiar with the right click dragging (used in Blender for example) but the mouse wheel provides the same functionnality and there is little chance to accidentally trigger it.

wjakob:
Hi Romain,

apologies for the delay -- I just took another look and I really like the spin arrows you added. Very nice!

It's almost ready to merge -- a few more more (hopefully quite minor) change requests:

1. The size of the spin buttons isn't considered in preferredSize(). If you rebase your code on the latest master, you can see that this causes layout issues in one widget in example 1. So if spinning is enabled, that should add the necessary amount to the preferred width.

2. The functions for registering the spin buttons are missing in ``python/python.cpp`` (should be easy -- you can just leave out the D() macro that fetches the documentation string, I'll add that later)

3. I would suggest to have the spin feature off by default (because who knows that meaningful bounds/increments are in a particular situation). To advertise the feature, it could be explicitly turned on for some of the widgets in example1/2.cpp.

3. Super nit-picky things for consistency in the codebase:

```
if(...)
```

should be

```
if (...)
````

Braces:
```
}
else
```

should be

```
} else
```

etc.

romp13:
Hi Wenzel,

No worries. Glad you like it :-)
I made the changes you requested. Only the documentation strings in py_doc.h are missing but you probably have a script to generate these.
By the way, I was wondering if you have some features/improvements todo list. I really like your library and would be happy to contribute when I can.

wjakob:
Hi Romain,

thanks, this looks all good! I've merged your PR and added the Python docstrings.
BTW: I made this change
https://github.com/wjakob/nanogui/commit/87168c0df85ad41094d476d51be9db20881c582f
which (amongst other things) re-adds a fix due to issue #71 that was removed in your commit. I assume that was not intentional.

Best,
Wenzel

Issue #67, opened 2016-05-17
Suggestion: Removing external libraries from this repo


jslee02:
I would love to use `nanogui` in my projects for graphical applications. As `nanogui` is, my project is also dependent on Eigen, and it makes a bit difficult to build them together. I understand that having all the external libraries in the same repo is very convenient when developing the library. However, it could cause dependency conflicts when the they have common dependencies.

To avoid this problem, I would like to suggest removing all the external libraries from this repo and changing `nanogui` to build with installed libraries using CMake `find_package`.

If this makes sense to you, then I'm willing to help the change. Please let me know what you think. Thanks!

wjakob:
I'd like to keep the dependencies there to make building the project absolutely straightforward. What I would accept is a PR which adds a flag to the CMakeLists.txt that can be used to specify a directory to an alternative Eigen directory that is used instead.

wjakob:
(Closing this issue, since there is no defect in the software.)

Issue #57, opened 2016-04-14
Remove dependencies?


fcvarela:
Hi, is there any reason this packs and builds against its own fork of (at least) glew and glfw? This is very common with ui toolkits (dearimgui notable exception), but it makes it near impossible to use them with any pre-existing app/renderer/engine/etc.

wjakob:
The reason is to make it trivial to drop into a new project and have a GUI from the get-go. Feel free to fork the project and remove them, but I plan to keep the main repository the way it is.

fcvarela:
Got it, makes sense. Thanks!

Issue #38, opened 2015-12-21
Add own event struct that exclude dependency from glfw in widgets


dalerank:
What you this about subj?
It may be need to add event converter from glfw event to custom event, but will do code clear and more portable. 

Thanks

wjakob:
Please feel free to fork nanogui and make such changes. However, I plan to keep the GLFW dependency for the main software package.

Issue #28, opened 2015-11-30
The framework is good but is to much tightly coupled very hard to embed it to other application 


meiry:
it is light and nice framework , but when trying to embed it to another application which already init glfw3 , there is need to mix the hosting app code  into Screen class ,   
this is pain , design as i see it and know should be flexible to add the framework into other .

wjakob:
Well, this is not a very helpful bug report. The Screen class specifically has support for working with an already GLFW context.

If you don't like it, you are welcome to submit improvements.

meiry:
Don't get me wrong , it is not bug only suggestion , 
the framework is great and very well written  but it is very hard to GUI layer and embed it .

wjakob:
Ok, I understand. Unfortunately my resources are very limited to work on suggestions. I do, however, accept patches if they contribute useful functionality or architectural improvements. Feel free to start a new pull request.

PR #72, opened 2016-06-16
Allowing different Eigen lib to be used

r-chris:
As suggested in #67 and #57 (#38, #28) it can make sense to reference your own external dependencies (e.g. Eigen library) instead of using the one provided in `ext/eigen`. This adds a `CMake` flag for the user to optionally enable a `find_path` search for `EIGEN_INCLUDE_DIR`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,6 +59,13 @@ endif()
 # Compile GLFW
 add_subdirectory("${CMAKE_CURRENT_SOURCE_DIR}/ext/glfw" "ext_build/glfw")
 
+if(NOT EIGEN_INCLUDE_DIR)
+    set(EIGEN_INCLUDE_DIR "ext/eigen")
+    list(APPEND NANOGUI_EXTRA_INCS
+        "${CMAKE_CURRENT_SOURCE_DIR}/${EIGEN_INCLUDE_DIR}"
+    )
+endif()
+
 # Python support: add NANOGUI_PYTHON flag to all targets
 if (NANOGUI_BUILD_PYTHON)
   list(APPEND NANOGUI_EXTRA_DEFS -DNANOGUI_PYTHON)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -381,7 +402,7 @@ else()
   # Create documentation for python plugin (optional target for developers)
   add_custom_target(mkdoc COMMAND
     python3 ${CMAKE_CURRENT_SOURCE_DIR}/ext/pybind11/tools/mkdoc.py
-    -Iinclude -Iext/nanovg/src -Iext/eigen -Iext/glfw/include
+    -Iinclude -Iext/nanovg/src -I${GLFW_INCLUDE_DIR} -I${GLFW_INCLUDE_DIR}
```
r-chris: just realized there is a copy&paste error here

r-chris:
I also added a `GLFW_INCLUDE_DIR` variable that would make a transition to the same paradigm for `GLFW` easy, but I am not sure if you agree to this and how the linking actually works?

```CMake
# Merge GLFW into the NanoGUI library
$<TARGET_OBJECTS:glfw_objects>
```

wjakob:
Sorry, I can't merge the PR in this form. The GLFW used by nanogui is based on a repository with build system modifications, so it's not straightforward to just use a different GLFW library. 

NANOGUI_BUILD_EIGEN doesn't seem like a good name since that is only a header library.

I would suggest to make the commit really minimal (i.e. 1-3 lines) and just not set EIGEN_INCLUDE_DIR when it was already defined by the parent project. The cache need not be involved to accomplish this.

r-chris:
No worries, this PR doesn't actually allow you to change GLFW to another directory. All it does in that respect is define an internal variable called `GLFW_INCLUDE_DIR` that the user can't change.

It wouldn't work for the general case to rely on a parent project to already set `EIGEN_INCLUDE_DIR`. As a user I might want to pull this repository in as a link-time dependency - at build time nanogui won't know anything about the project that wants to link with it.

r-chris:
I removed the `GLFW` related reference and followed your advice to remove `NANOGUI_BUILD_EIGEN` after all. What I didn't realize is how easily I can call `ccmake ../nanogui/ -DEIGEN_INCLUDE_DIR=/usr/local/include/eigen3` to build nanogui independently for any project.

wjakob:
two more requests: can you preface the EIGEN_ variable with NANOGUI_? (namespacing rules). Also, it looks like there is now a merge conflict. I will merge it after this.

Thank you,
Wenzel

r-chris:
Sure, with the previous commit that was probably expected.

PR #74, opened 2016-06-16
Adding install targets to nanogui

r-chris:
I couldn't figure out if you have purposely removed any `make install` targets or never actually used them (even though #1 appears to reference installing things). Regardless, I am using `ExternalProject_Add` and it complains that no install targets are available - so I added them here - in case you are interested to merge this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,7 @@ option(NANOGUI_BUILD_EXAMPLE "Build NanoGUI example application?" ON)
 option(NANOGUI_BUILD_SHARED  "Build NanoGUI as a shared library?" ON)
 option(NANOGUI_BUILD_PYTHON  "Build a Python plugin for NanoGUI?" ON)
 option(NANOGUI_USE_GLAD      "Build a Python plugin for NanoGUI?" ${NANOGUI_USE_GLAD_DEFAULT})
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -235,6 +235,17 @@ if (NANOGUI_BUILD_SHARED)
...
 endif()
 
+install(
+  TARGETS nanogui
+  LIBRARY DESTINATION lib
+  ARCHIVE DESTINATION lib
+  )
+
+install(
+  DIRECTORY include/nanogui DESTINATION include
```
wjakob: Do these files get installed to <destination>/include or <destination>/include/nanogui? The latter is what would be intended by the library.

wjakob:
Can you wrap all of the install targets into "if (NANOGUI_INSTALL)" statements? That parameter should be an option at the top with default value ON.

CMakeLists.txt
r-chris: It is the latter version. The folder `nanogui` will be installed into the `DESTINATION/include` folder, so that you can include like this `#include<nanogui/nanogui.h>`

r-chris:
Alright, changed it according to your request.

wjakob:
LGTM, thanks!

PR #83, opened 2016-07-15
Adding pre- and post- process methods to Screen that get called befor…

sjm324:
…e / after the drawAll method respectively.  Allows for communicating with other hardware on a per-frame basis that GLFW does not support, if needed.  Otherwise, empty stubs in a void method should have minimal to no impact on performance depending on optimization levels / compiler.

This may not be appropriate to bring into the master, but in order to coordinate per-frame manipulations of viewing the scene with hardware GLFW doesn't support (they only do keyboard and mouse and some gamepads...reasonably so) I needed a method before / after `drawAll` to communicate with the device.

I figured I would submit a pull request since I already implemented, but will close it if you don't think it fits in.  If the functionality is desirable, but the execution needs to change that can be done.  The only other option I could think of was giving the `mainloop` a functor or something, which seemed kind of obtuse.

I put some print statements (not in the commit) in really quickly to test it and the cpp and python examples seem to work as expected.


wjakob:
Since drawAll is a virtual method, the user can override it and do these calls himself at the beginning or the end. In that sense, it doesn't seem to be the case that this adds functionality that wasn't there before.

sjm324:
:faceslap: you are correct. Has that method always been exposed?!

Anyway, that's a much better solution...sorry!

Issue #85, opened 2016-07-20
(unix) strip command fails when CMAKE_LIBRARY_OUTPUT_DIRECTORY is set


svenevs:
I don't understand what this is doing / why

```
if (NANOGUI_BUILD_SHARED AND NOT ${U_CMAKE_BUILD_TYPE} MATCHES DEBUG)
  if (APPLE)
    # Strip .dylib library on OSX
    add_custom_command(TARGET nanogui POST_BUILD COMMAND strip -u -r ${CMAKE_CURRENT_BINARY_DIR}/libnanogui.dylib)
  elseif(UNIX)
    # Strip .so library on Linux
    add_custom_command(TARGET nanogui POST_BUILD COMMAND strip ${CMAKE_CURRENT_BINARY_DIR}/libnanogui.so)
  endif()
endif()
```

I would prefer to continue using a shared library Release build, but the circumstances that be require I set the `CMAKE_LIBRARY_OUTPUT_DIRECTORY`.  So since `libnanogui.*` will get put there rather than `CMAKE_CURRENT_BINARY_DIR`, `strip` will fail.

Does removing the library suffix have a specific purpose?  Would something like the following inside the `NANOGUI_BUILD_SHARED AND NOT ...` section work?

```
get_property(NANOGUI_ALTERNATE_LIBRARY_LOCATION GLOBAL PROPERTY LIBRARY_OUTPUT_DIRECTORY)
if(NANOGUI_ALTERNATE_LIBRARY_LOCATION)
  set(NANOGUI_LIBRARY_DIRECTORY "${CMAKE_LIBRARY_OUTPUT_DIRECTORY}"
else()
  set(NANOGUI_LIBRARY_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}")
endif()
```

and then use `${NANOGUI_LIBRARY_DIRECTORY}/libnanogui...` in each respective branch?

wjakob:
The strip command reduces the size of the generated executable. I'm open to contributions which rewrite the add_custom_command invocations using CMake generator expressions so that it can find the library wherever it ends up. I don't think it's a good idea to introduce a new property for that.

svenevs:
I believe the above PR fixes this, the `CMAKE_*_OUTPUT_DIRECTORY` variables are somewhat under-documented and/or not used consistently throughout the community.  My understanding is nobody building a library should ever be setting these properties, but I didn't realize `get_property` did what it does.

[...]

PR #87, opened 2016-07-23
support for CMAKE_LIBRARY_OUTPUT_DIRECTORY alternate output directory…

svenevs:
… on shared library build

Supports alternate library output paths specified, resolves https://github.com/wjakob/nanogui/issues/85

I made a quick example repository to build both:

https://github.com/svenevs/nanogui-strip-test

You should be able to just clone `--recursive` and execute `compile_both.sh`.  It works on OSX, but I am currently traveling and unable to check if it works on Linux...though I don't see why that would fail.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -286,12 +286,13 @@ if (NANOGUI_BUILD_SHARED)
 endif()
 
 if (NANOGUI_BUILD_SHARED AND NOT ${U_CMAKE_BUILD_TYPE} MATCHES DEBUG)
+  # Platform-specific strip flags for reducing the library size.
   if (APPLE)
     # Strip .dylib library on OSX
-    add_custom_command(TARGET nanogui POST_BUILD COMMAND strip -u -r ${CMAKE_CURRENT_BINARY_DIR}/libnanogui.dylib)
+    add_custom_command(TARGET nanogui POST_BUILD COMMAND strip -u -r "$<TARGET_FILE_DIR:nanogui>/$<TARGET_FILE_NAME:nanogui>")
   elseif(UNIX)
     # Strip .so library on Linux
-    add_custom_command(TARGET nanogui POST_BUILD COMMAND strip ${CMAKE_CURRENT_BINARY_DIR}/libnanogui.so)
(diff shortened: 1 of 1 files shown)


wjakob:
I think it would be much nicer to do this via generator expressions. (https://cmake.org/cmake/help/v3.3/manual/cmake-generator-expressions.7.html) rather than defining yet another parameter.

wjakob:
..specifically ``$<TARGET_FILE_DIR:tgt>``. That way, the "exterior" project could modify the library path, and the cmake could would still find it in the right place. Can you please modify the PR accordingly?

svenevs:
Oh these are very nice!

I wasn't able to get it working with `TARGET_FILE_DIR`, or `TARGET_FILE_NAME` on their own though, I had to do

```
if (NANOGUI_BUILD_SHARED AND NOT ${U_CMAKE_BUILD_TYPE} MATCHES DEBUG)
  # Platform-specific strip flags / library suffix for reducing the library size.
  if (APPLE)
    # Strip .dylib library on OSX
    add_custom_command(TARGET nanogui POST_BUILD COMMAND strip -u -r "$<TARGET_FILE_DIR:nanogui>/$<TARGET_FILE_NAME:nanogui>")
  elseif(UNIX)
    # Strip .so library on Linux
    add_custom_command(TARGET nanogui POST_BUILD COMMAND strip "$<TARGET_FILE_NAME:nanogui>/$<TARGET_FILE_NAME:nanogui>")
  endif()
endif()
```

Am I using these incorrectly?  I would think only the `FILE_NAME` should be needed.  Similarly, I was able to get it working the same way if I used `TARGET_LINKER_FILE_{DIR/FILE}` in the same fashion.  Both work even if I set `CMAKE_RUNTIME_OUTPUT_DIRECTORY` to something different.

I think `TARGET_FILE_{DIR,NAME}` are more appropriate, since for the nanogui CMake, that is the target, yes?

wjakob:
This looks good to me! Can you squash your changes into a single commit?

Thanks,
Wenzel

svenevs:
Squashed

wjakob:
Thanks!

PR #89, opened 2016-07-25
Add example3 c++ to demonstrate the usage of nanogui while also managing glfw independently.

AlwaysGeeky:
I added the example3.cpp to start the ball rolling for helping to improve the usage of nanogui when you want to also manage glfw and the general windows loop yourself.

It seems this is something the nanogui project is in need of and also needs to be improved since I personally see this as the main use-case of anyone wanting to take nanogui and use it themselves. I dont really think many people will would want to take nanogui and drop it into their projects for the gui functionality but at the cost of giving up and handing over all management and running of the glfw context and windows management to nanogui.


src/screen.cpp
```diff
@@ -72,7 +72,7 @@ static float get_pixel_ratio(GLFWwindow *window) {
 }
 Screen::Screen()
     : Widget(nullptr), mGLFWWindow(nullptr), mNVGContext(nullptr),
-      mCursor(Cursor::Arrow), mBackground(0.3f, 0.3f, 0.32f),
+      mCursor(Cursor::Arrow), mBackground(0.3f, 0.3f, 0.32f), mCaption(""),
```
wjakob: Not sure what this change accomplishes? std::string is empty by default. (?)

wjakob:
Thank you very much! I'm sure that this will be very helpful to others integrating NanoGUI into their projects. Two comments:

1. Could you squash this into 2 commits? (1 for the changes/fixes in nanogui, and one for your new example3)

2. Your code uses tabs, but NanoGUI uses spaces for indentation. Please switch your commits to spaces :)

Thanks,
Wenzel

src/screen.cpp
AlwaysGeeky: Ahhh you are right, I was being over zealous, since I thought some un-initialized params in the constructor were causing issues. I will remove this.

AlwaysGeeky:
In terms of splitting the pull request... I'm not really a pro expert on git, but I have forked the code and commited the changes in my branch. Is there a way I can customize the pulls or do some pull request magic to get the desired result you are looking for?

wjakob:
never mind, I can do it too. (I am traveling and probably won't get to it until next week though.)

AlwaysGeeky:
I suppose I could just delete this fork and pull request and do it that way, if that would allow you to merge it easier and quicker... I'll do that now, since it is only a couple of minor changes.

PR #91, opened 2016-07-25
Fix screen for applications that want to manage their own glfw and not rely on nanogui

AlwaysGeeky:
Also include a new example, example3.cpp that demonstrates a way to create your own windows management code with glfw but still create GUI objects with nanogui.

### DIFF
--- a/include/nanogui/screen.h
+++ b/include/nanogui/screen.h
@@ -192,7 +192,7 @@ protected:
     std::string mCaption;
     bool mShutdownGLFWOnDestruct;
     bool mFullscreen;
-	bool mOwningGLFWContext;
+    bool mOwningGLFWContext;
 };
 
 NAMESPACE_END(nanogui)
--- a/include/nanogui/screen.h
+++ b/include/nanogui/screen.h
@@ -192,6 +192,7 @@ protected:
     std::string mCaption;
     bool mShutdownGLFWOnDestruct;
     bool mFullscreen;
+	bool mOwningGLFWContext;
 };
 
 NAMESPACE_END(nanogui)
--- a/src/example3.cpp
+++ b/src/example3.cpp
@@ -79,12 +79,12 @@ int main(int /* argc */, char ** /* argv */) {
     FormHelper *gui = new FormHelper(screen);
     ref<Window> nanoguiWindow = gui->addWindow(Eigen::Vector2i(10, 10), "Form helper example");
     gui->addGroup("Basic types");
-    gui->addVariable("bool", bvar);
+    gui->addVariable("bool", bvar)->setTooltip("Test tooltip.");
     gui->addVariable("string", strval);
 
     gui->addGroup("Validating fields");
     gui->addVariable("int", ivar)->setSpinnable(true);
-    gui->addVariable("float", fvar);
+    gui->addVariable("float", fvar)->setTooltip("Test.");
     gui->addVariable("double", dvar)->setSpinnable(true);
 
     gui->addGroup("Complex types");
@@ -92,7 +92,7 @@ int main(int /* argc */, char ** /* argv */) {
     gui->addVariable("Color", colval);
 
     gui->addGroup("Other widgets");
-    gui->addButton("A button", []() { std::cout << "Button pressed." << std::endl; });
+    gui->addButton("A button", []() { std::cout << "Button pressed." << std::endl; })->setTooltip("Testing a much longer tooltip, that will wrap around to new lines multiple times.");;
 
     screen->setVisible(true);
     screen->performLayout();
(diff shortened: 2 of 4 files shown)


include/nanogui/screen.h
```diff
@@ -192,6 +192,7 @@ class NANOGUI_EXPORT Screen : public Widget {
     std::string mCaption;
     bool mShutdownGLFWOnDestruct;
     bool mFullscreen;
+	bool mOwningGLFWContext;
```
wjakob: Indentation issue (are you using tabs? NanoGUI uses only spaces.)

src/screen.cpp
```diff
@@ -300,10 +309,12 @@ void Screen::setVisible(bool visible) {
     if (mVisible != visible) {
         mVisible = visible;
 
-        if (visible)
-            glfwShowWindow(mGLFWWindow);
-        else
-            glfwHideWindow(mGLFWWindow);
+		if (mOwningGLFWContext) {
```
wjakob: tab indentation issue

src/screen.cpp
```diff
@@ -368,20 +379,27 @@ void Screen::drawWidgets() {
...
-            nvgTextAlign(mNVGContext, NVG_ALIGN_CENTER | NVG_ALIGN_TOP);
+            nvgTextAlign(mNVGContext, NVG_ALIGN_LEFT | NVG_ALIGN_TOP);
             nvgTextLineHeight(mNVGContext, 1.1f);
             Vector2i pos = widget->absolutePosition() +
                            Vector2i(widget->width() / 2, widget->height() + 10);
 
-            nvgTextBoxBounds(mNVGContext, pos.x(), pos.y(), tooltipWidth,
+			nvgTextBounds(mNVGContext, pos.x(), pos.y(),
                              widget->tooltip().c_str(), nullptr, bounds);
-
```
wjakob: ditto -- please do a pass over all the changes and replace tabs with four spaces ;)

src/screen.cpp
```diff
@@ -300,10 +309,12 @@ void Screen::setVisible(bool visible) {
     if (mVisible != visible) {
         mVisible = visible;
 
-        if (visible)
-            glfwShowWindow(mGLFWWindow);
-        else
-            glfwHideWindow(mGLFWWindow);
+        if (mOwningGLFWContext) {
```
wjakob: Why disable setVisible() for non-owned GLFW context? I think it would be quite reasonable to leave this on.

src/screen.cpp
```diff
@@ -368,20 +379,27 @@ void Screen::drawWidgets() {
...
             Vector2i pos = widget->absolutePosition() +
                            Vector2i(widget->width() / 2, widget->height() + 10);
 
-            nvgTextBoxBounds(mNVGContext, pos.x(), pos.y(), tooltipWidth,
-                             widget->tooltip().c_str(), nullptr, bounds);
+            nvgTextBounds(mNVGContext, pos.x(), pos.y(),
+                            widget->tooltip().c_str(), nullptr, bounds);
+            int h = (bounds[2] - bounds[0]) / 2;
+            if (h > tooltipWidth / 2)
+            {
```
wjakob: Super-minor: brace should be on previous line.

src/example3.cpp
```diff
@@ -0,0 +1,163 @@
...
+    gui->addGroup("Other widgets");
+    gui->addButton("A button", []() { std::cout << "Button pressed." << std::endl; })->setTooltip("Testing a much longer tooltip, that will wrap around to new lines multiple times.");;
+
+    screen->setVisible(true);
+    screen->performLayout();
+    nanoguiWindow->center();
+
+
+    glfwSetCursorPosCallback(window,
+        [](GLFWwindow *w, double x, double y) {
```
wjakob: Super-minor: lambda function body and the following brace should be indented by another 4 spaces here and below.

wjakob:
I added a few more comments -- after that, it looks good to me.

src/screen.cpp
AlwaysGeeky: If I am managing my own GLFW and windowing code, and just using a pointer to a 'screen' object for my GUI, having the whole window hidden just because I call: screen->setVisible(false) is misleading, what if I just want to hide the GUI alone, but keep my true window visible?

wjakob: Well, ``screen`` is a wrapper around the GLFW window, so ``setVisible()`` should be forwarded to GLFW. Whether the Screen also owns the GLFW window and is responsible for construction/deconstruction should not change the behavior of this function. Hiding the individual windows can be accomplished with their respective setVisible() functions.

wjakob:
(Please squash everything into one commit when you're done.)

wjakob:
merged, thanks!

PR #97, opened 2016-08-04
New Image View Widget

Thorius94:
Removed the old image view widget and replaced it with a class with
a new design. This new class uses OpenGL textures directly and allows
the user to scale and move the image.

Updated all occurences appropriately (both in C++ and in Python).
Updated example 1 to use the new widget.

Also, fixed 2 typos in README and added VS code folder to .gitignore


wjakob:
This looks great, thank you!

PR #101, opened 2016-08-06
refresh_thread minor fix

xSaiteix:
Now mainLoop can correctly handle 0 refresh-time (mainLoop(0))


wjakob:
Actually having a refresh time == 0 is probably a really bad idea since the event queue will be spammed with dummy events. The right thing will be to update the documentation and say that a positive value is needed.

PR #104, opened 2016-08-09
only account for header height if Window has a title

selfshadow:
This logic is in GroupLayout, but absent from some other layouts. Without this check, there can be padding if, for instance, one uses a GridLayout with a Popup.

Note: I have not tested all layouts in all configurations. For instance, I'm not sure if the behaviour makes sense for Orientation::Vertical.

Perhaps there is also room for simplification, by moving some of this logic to a helper function or member?


wjakob:
Looks good, thanks! I think it make sense for all configurations.

PR #105, opened 2016-08-11
fixed Screen::setSize high-DPI behaviour on windows

selfshadow:
Previously, calling setSize() on a high-DPI display would lead to the window being too small.

Unresolved: should setSize() do nothing if mFullscreen is true?


wjakob:
Hi Steve :)

This looks good, thanks!

PR #106, opened 2016-08-11
fixed window size when minimising on windows

### LINKED ISSUE #95: Window not restoring properly after minimization
This happens with the examples that come with the library: if I minimize then restore the main window, its size does not restore to the proper thing. It comes back up with a height of 0 and a width just large enough for the title to show.

I think you're not calling glfwSetWindowIconifyCallback anywhere. I just wanted to bring it to your attention and see if you had a fix planned, in the meantime we're keeping track of our window size manually.
Thanks!

selfshadow:
When the window is minimised, glfwGetWindowSize() returns {0, 0} for the size on Windows, since it uses GetClientRect() internally. This leads to a division by zero when calculating mPixelRatio, which later triggers the window size to be updated in Screen::drawWidgets():

```
#if defined(_WIN32)
    if (mPixelRatio != newPixelRatio && !mFullscreen)
        glfwSetWindowSize(mGLFWWindow, mSize.x() * newPixelRatio / mPixelRatio, mSize.y() * newPixelRatio / mPixelRatio);
#endif
```

Fixes #95

### DIFF
--- a/src/layout.cpp
+++ b/src/layout.cpp
@@ -30,7 +30,8 @@ Vector2i BoxLayout::preferredSize(NVGcontext *ctx, const Widget *widget) const {
     Vector2i size = Vector2i::Constant(2*mMargin);
 
     int yOffset = 0;
-    if (dynamic_cast<const Window *>(widget)) {
+    const Window *window = dynamic_cast<const Window *>(widget);
+    if (window && !window->title().empty()) {
         if (mOrientation == Orientation::Vertical)
             size[1] += widget->theme()->mWindowHeaderHeight - mMargin/2;
         else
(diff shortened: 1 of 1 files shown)


wjakob:
Quick question: wouldn't it be better to *not* update mPixelRatio in this case? (rather than setting it to 1.0, which could be incorrect)

selfshadow:
Yes, I was starting to question this myself, particularly as I'd neglected to check the behaviour with a high-DPI display. I'll take a look and update the PR accordingly.

selfshadow:
I appear to have made a mess of this branch. I'll try to fix it so there's just the relevant change.

selfshadow:
I wasn't expecting GitHub to auto-close the PR when I removed the old commits. *sigh*

PR #107, opened 2016-08-11
fixed window size when minimising on windows

### LINKED ISSUE #95: Window not restoring properly after minimization

selfshadow:
When the window is minimised, glfwGetWindowSize() returns {0, 0} for the size on Windows, since it uses GetClientRect() internally. This leads to a division by zero when calculating mPixelRatio, which later triggers the window size to be updated in Screen::drawWidgets():
```
#if defined(_WIN32)
    if (mPixelRatio != newPixelRatio && !mFullscreen)
        glfwSetWindowSize(mGLFWWindow, mSize.x() * newPixelRatio / mPixelRatio, mSize.y() * newPixelRatio / mPixelRatio);
#endif
```

Fixes #95

### DIFF
--- a/src/screen.cpp
+++ b/src/screen.cpp
@@ -369,7 +369,8 @@ void Screen::drawWidgets() {
     glViewport(0, 0, mFBSize[0], mFBSize[1]);
 
     /* Calculate pixel ratio for hi-dpi devices. */
-    mPixelRatio = (float) mFBSize[0] / (float) mSize[0];
+    if (mSize[0])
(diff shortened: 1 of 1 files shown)


wjakob:
Great, thanks!

PR #108, opened 2016-08-11
Fixes scissors not working correctly when multiple scissors are within the same widget

willxinc:
Use nvgIntersectScissor instead of nvgScissor.

nvgSave(ctx) before any new scissors
and 
nvgRestore(ctx) after and scissor resets.


wjakob:
That's a great catch -- thank you!

PR #109, opened 2016-08-11
Vscrollpanel popup follow fix

willxinc:
popup.cpp#L65 to 67 are questionable, but the anchor height always seems to be a fixed value (previously).

Adjusting this changes the offset of the anchor (which we update when the vscrollpanel is scrolled)

### DIFF
--- a/include/nanogui/vscrollpanel.h
+++ b/include/nanogui/vscrollpanel.h
@@ -13,6 +13,7 @@
 #pragma once
 
 #include <nanogui/widget.h>
+#include <nanogui/popup.h>
 
 NAMESPACE_BEGIN(nanogui)
 
@@ -27,6 +28,7 @@ public:
     virtual bool mouseButtonEvent(const Vector2i &p, int button, bool down, int modifiers) override;
     virtual bool mouseMotionEvent(const Vector2i &p, const Vector2i &rel, int button, int modifiers) override;
     virtual void draw(NVGcontext *ctx) override;
+	virtual float getOffset();
     virtual void save(Serializer &s) const override;
     virtual bool load(Serializer &s) override;
 protected:
--- a/include/nanogui/vscrollpanel.h
+++ b/include/nanogui/vscrollpanel.h
@@ -27,7 +27,7 @@ public:
     virtual bool mouseButtonEvent(const Vector2i &p, int button, bool down, int modifiers) override;
     virtual bool mouseMotionEvent(const Vector2i &p, const Vector2i &rel, int button, int modifiers) override;
     virtual void draw(NVGcontext *ctx) override;
-	virtual float getOffset();
+    virtual float getOffset();
     virtual void save(Serializer &s) const override;
     virtual bool load(Serializer &s) override;
 protected:
--- a/include/nanogui/vscrollpanel.h
+++ b/include/nanogui/vscrollpanel.h
@@ -13,7 +13,6 @@
 #pragma once
 
 #include <nanogui/widget.h>
-#include <nanogui/popup.h>
 
 NAMESPACE_BEGIN(nanogui)
 
--- a/src/popup.cpp
+++ b/src/popup.cpp
@@ -35,7 +35,10 @@ void Popup::performLayout(NVGcontext *ctx) {
 void Popup::refreshRelativePlacement() {
     mParentWindow->refreshRelativePlacement();
     mVisible &= mParentWindow->visibleRecursive();
-    mPos = mParentWindow->position() + mAnchorPos - Vector2i(0, mAnchorHeight - mParentPanel->getOffset());
+    mPos = mParentWindow->position() + mAnchorPos - Vector2i(0, mAnchorHeight);
+	if (mParentPanel){
+		mPos += Vector2i(0, mParentPanel->getOffset());
+	}
 }
 
 void Popup::draw(NVGcontext* ctx) {
--- a/src/popup.cpp
+++ b/src/popup.cpp
@@ -19,7 +19,7 @@ NAMESPACE_BEGIN(nanogui)
     
 Popup::Popup(Widget *parent, Window *parentWindow)
     : Window(parent, ""), mParentWindow(parentWindow),
-      mAnchorPos(Vector2i::Zero()), mAnchorHeight(30) {
+      mAnchorPos(Vector2i::Zero()), mAnchorHeight(30), mParentPanel(nullptr){
 }
 
 void Popup::performLayout(NVGcontext *ctx) {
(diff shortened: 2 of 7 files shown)


include/nanogui/vscrollpanel.h
```diff
@@ -27,6 +28,7 @@ class NANOGUI_EXPORT VScrollPanel : public Widget {
     virtual bool mouseButtonEvent(const Vector2i &p, int button, bool down, int modifiers) override;
     virtual bool mouseMotionEvent(const Vector2i &p, const Vector2i &rel, int button, int modifiers) override;
     virtual void draw(NVGcontext *ctx) override;
+	virtual float getOffset();
```
wjakob: Indentation issue here and elsewhere: spaces, not tabs

wjakob:
I don't have enough context to understand this PR. Is there a specific issue that this addresses? (e.g. something with vscrollpanel that contains popup buttons?)

willxinc:
When you put popup buttons within a vscrollpanel and activate the popup, the popup that is created's location is where it would be, if you never scrolled at all.

ie. if you scroll the panel down 200px, then click on a button that was 200px down, the popup will be created, not next to the button, but 200px down from it.

I can provide an example if you'd like.

wjakob:
If it's not too much trouble, that would be great (also to ensure that this doesn't break in the future) -- something very basic will do.

willxinc:
[Example](https://gist.github.com/willxinc/53b77ccb2648050cfe2d1ef72df522fc)

Click on popups 0-6 without scrolling the window, notice that they're in the correct position.

Leave any of them on, and scroll the window, noticing that the popups stay in place.

After scrolling the window, click on any of the popup buttons from 7-14, notice that they are drawn as if the window hasn't been scrolled.

src/popup.cpp
```diff
@@ -62,9 +62,9 @@ void Popup::draw(NVGcontext* ctx) {
     nvgBeginPath(ctx);
     nvgRoundedRect(ctx, mPos.x(), mPos.y(), mSize.x(), mSize.y(), cr);
 
-    nvgMoveTo(ctx, mPos.x()-15,mPos.y()+mAnchorHeight);
```
wjakob: This is drawing the little triangle AFAIK. Why is usage of mAnchorHeight suspect here?

wjakob: (there are some projects that use different anchor heights.)

willxinc:
Ah, in that case another member would need to be added that contains the scroll offset instead of re-purposing the mAnchorHeight member.

src/vscrollpanel.cpp
```diff
@@ -81,11 +81,24 @@ void VScrollPanel::draw(NVGcontext *ctx) {
...
+    float offset = -mScroll*(mChildPreferredHeight - mSize.y());
 
     nvgSave(ctx);
     nvgTranslate(ctx, mPos.x(), mPos.y());
     nvgIntersectScissor(ctx, 0, 0, mSize.x(), mSize.y());
-    nvgTranslate(ctx, 0, -mScroll*(mChildPreferredHeight - mSize.y()));
+    nvgTranslate(ctx, 0, offset);
+
+
+    Widget *parentWindow = (Widget*)window();
```
wjakob: Why not directly try to cast the parent window into a ``Popup``? I don't understand the need for a loop here.

willxinc: The loop applies the offset to all the popups that are a child of the current screen.
( This is because when you make a popup button, the popup get's attached to the screen since it's a window ).
This is also why the nvgTranslate doesn't work on the popup, hence the workaround using the anchor.

wjakob: I'm not a big fan of the loop over child widgets. What about redesigning this patch so that your change is done in the popup button class, specifically line 66? This is the place where you could auto-detect whether one of the parents is a ScrollPanel and if so, adjust the position.

willxinc:
I've made another commit with an alternate solution, I still feel like there could be a better solution however this shouldn't break anything.

src/popupbutton.cpp
```diff
@@ -63,6 +63,15 @@ void PopupButton::performLayout(NVGcontext *ctx) {
 
     const Window *parentWindow = window();
 
+	Widget* checkScroll = parent();
```
wjakob: incorrect indentation.

src/popupbutton.cpp
```diff
@@ -63,6 +63,15 @@ void PopupButton::performLayout(NVGcontext *ctx) {
...
+	while (checkScroll != parentWindow){
```
wjakob: rather than checking for equality with parentWindow, do a loop until ``widget`` (rename variable to this) becomes ``nullptr``.

src/popupbutton.cpp
```diff
@@ -63,6 +63,15 @@ void PopupButton::performLayout(NVGcontext *ctx) {
...
+		if (VScrollPanel* derived = dynamic_cast<VScrollPanel*>(checkScroll)){
+			mPopup->setParentPanel(derived);
```
wjakob: Rather than introducing a new ``setParentPanel`` call here, why not issue a modified version of the ``mPopup->setAnchorPos()`` call from below which takes into account the scroll panel offset?

wjakob:
I'd like to commit a nice solution, so let's refine this a bit. I added some comments to your latest commit.

willxinc:
setAnchorPos works, but then the drawing of the triangles become offset (as well as overwrite any anchors that the user may have setup)
Also this doesn't update the popup position as you scroll, since you are only calling performLayout once.

VirgiliuC:
Hi

VScrollPanel doesn't seem to behave correctly with any controls, when planted into a peer-based layout (that is, not straight into a window, but into a hierarchy, with layout siblings). When you scroll, the controls are no longer getting the correct click position.

I have created a new (simple) widget that wraps around VScrollpanel, that **does work** correctly with the controls (I haven't changed the vscrollpanel's original code). 
As I wrote it about 2-3 months ago, I cannot exactly remember now about the popups, but my feeling is that it works fine like all the others (I don't have my code with me right now). I am planning next week to do a fork and commit/propose that widget, as I am also working on other widgets that may be useful for the framework (actually I had to fix that vscroll thing as a mandatory task for my new composite ngui widgets).

Virgiliu

wjakob:
@VirgiliuC please try with the latest version, there was a patch yesterday

VirgiliuC:
Will do so, thanks Wenzel.

willxinc:
Before the latest patch, you would need to wrap vscroll in another widget in order to get correct click locations.
This does not fix popups, as popups are not attached to the widget, but to the screen directly (since they are windows)

wjakob:
fixed, see 37ae0c35978a1a978131f4427bdc539a083a7a5a

PR #112, opened 2016-08-19
Added margin initialization with default parameter to AdvancedGridLayout constructor

avasilkov:
mMargin value was not initialized anywhere in AdvancedGridLayout. 
Added default margin parameter to AdvancedGridLayout constructor with 0 value like in GridLayout.
If user didn't call setMargin(value) explicitly after AdvancedGridLayout creation, mMargin had undefined value and because of it, positions of widgets computed by performLayout were very wrong.

### DIFF
--- a/include/nanogui/layout.h
+++ b/include/nanogui/layout.h
@@ -248,7 +248,7 @@ public:
         }
     };
 
-    AdvancedGridLayout(const std::vector<int> &cols = {}, const std::vector<int> &rows = {});
+	AdvancedGridLayout(const std::vector<int> &cols = {}, const std::vector<int> &rows = {}, int margin = 0);
 
     int margin() const { return mMargin; }
     void setMargin(int margin) { mMargin = margin; }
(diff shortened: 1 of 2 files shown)


include/nanogui/layout.h
```diff
@@ -248,7 +248,7 @@ class NANOGUI_EXPORT AdvancedGridLayout : public Layout {
         }
     };
 
-    AdvancedGridLayout(const std::vector<int> &cols = {}, const std::vector<int> &rows = {});
+	AdvancedGridLayout(const std::vector<int> &cols = {}, const std::vector<int> &rows = {}, int margin = 0);
```
wjakob: spaces, not tabs

wjakob:
Good catch, just a minor formatting issue left. Can you fix that and squash into one commit?

avasilkov:
@wjakob Is it okay if I resubmit this pull request completely? I will delete the fork and do everything again, because I broke something locally and now there is no changes to commit and there is no tab to find either.

PR #113, opened 2016-08-19
Added margin initialization with default parameter to AdvancedGridLayout constructor(no tabs)

avasilkov:
mMargin value was not initialized anywhere in AdvancedGridLayout. 
Added default margin parameter to AdvancedGridLayout constructor with 0 value like in GridLayout.
If user didn't call setMargin(value) explicitly after AdvancedGridLayout creation, mMargin had undefined value and because of it, positions of widgets computed by performLayout were very wrong.
//version without tabs.

### DIFF
--- a/include/nanogui/layout.h
+++ b/include/nanogui/layout.h
@@ -248,7 +248,7 @@ public:
         }
     };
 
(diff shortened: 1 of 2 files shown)


wjakob:
LGTM, thanks!

PR #114, opened 2016-08-20
Adding option to popup to appear at right or left side

avasilkov:
It looks like this: http://imgur.com/xc49tzo
and to use it you need to call `popupBtn->setPopupSide(PopupSide::LEFTSIDE);` after popupButton creation.
Not sure about save/load changes, can I save it as enum or need to convert to int like I'm doing?
And if a user calls setPopupSide on popup window instead of a button, then mChevronIcon will not be set to opposite side value.

### DIFF
--- a/include/nanogui/popup.h
+++ b/include/nanogui/popup.h
@@ -16,6 +16,10 @@
 
 NAMESPACE_BEGIN(nanogui)
 
+enum class PopupSide{
+    LEFTSIDE=0, RIGHTSIDE
+};
+
 /**
  * \brief Popup window for combo boxes, popup buttons, nested dialogs etc.
  *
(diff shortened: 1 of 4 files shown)


svenevs:
For the serializing, you can explicitly declare the size of an enum class:
```
enum class PopupSide : unsigned char { ... };
```
More info here: http://www.cprogramming.com/c++11/c++11-nullptr-strongly-typed-enum-class.html

I think you can even avoid casting to a char when you feed it to the serializer, but I'm not sure.

zivil:
I really like the style of the toolkit. But I want to ask do the richtext widget will join the lib?

wjakob:
@zivil: when you implement it and submit a PR ;)

zivil:
@wjakob Maybe I will hava a try when I am free.

PR #115, opened 2016-08-27
Fixed crashing if no text is pasted in a textbox

willxinc:
More specifically, if non-text is pasted into a textbox, an exception is thrown.

ie. pasting a file, or an image


wjakob:
Can you please squash these into a single commit?

PR #118, opened 2016-08-31
Adding minimize window method to the window widget

avasilkov:
How to use:
```
window->toggleMinimized();
window->setMinimized(true/false);
```

When minimized window looks like a window without content with height of a header that contains title or button panel or 0 if title is empty and button panel doesn't exist or not visible.

On minimize I'm saving visible state for each widget in mChildren inside unordered_map and hide all children with setVisible(false) except button panel(optional). I also store current size before minimization. I set size to current width and mTheme->mWindowHeaderHeight

On setMinimize(false) I'm restoring saved size and for each widget I'm restoring its visible state except those that were set to be true while being minimized or new widgets.

Convenient use case:
```
        Button* btn = window->buttonPanel()->add<Button>("", ENTYPO_ICON_CHEVRON_SMALL_UP);
        btn->setCallback([window, btn](){
            window->toggleMinimized();
            if(window->minimized()){
                btn->setIcon(ENTYPO_ICON_CHEVRON_SMALL_DOWN);
            }else{
                btn->setIcon(ENTYPO_ICON_CHEVRON_SMALL_UP); 
            }
        });
```

Not sure about method names, variable names and whether I should include something into save/load methods.

### DIFF
--- a/include/nanogui/window.h
+++ b/include/nanogui/window.h
@@ -12,6 +12,7 @@
 #pragma once
 
 #include <nanogui/widget.h>
+#include <unordered_map>
 
 NAMESPACE_BEGIN(nanogui)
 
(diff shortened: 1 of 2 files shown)


wjakob:
@avasilkov This is a reasonable change, but I don't think it needs to go into the core library. I am thus closing this PR.

PR #123, opened 2016-09-04
added out/in knob radius koeffs to slider

dalerank:


wjakob:
Thank you for your contribution. I don't think this needs to go into NanoGUI though.

PR #127, opened 2016-09-13
Fix: Example3 calls glViewport too early.

lmurmann:
In example3, glViewport was called before initializing GLAD, causing the example to crash on Windows. Just removing the glViewport call seems to work fine.

### DIFF
--- a/src/example3.cpp
+++ b/src/example3.cpp
@@ -81,7 +81,7 @@ int main(int /* argc */, char ** /* argv */) {
 
     int width, height;
     glfwGetFramebufferSize(window, &width, &height);
-    glViewport(0, 0, width, height);
+
     glfwSwapInterval(0);
     glfwSwapBuffers(window);
 
--- a/src/example3.cpp
+++ b/src/example3.cpp
@@ -79,15 +79,15 @@ int main(int /* argc */, char ** /* argv */) {
     }
     glfwMakeContextCurrent(window);
 
-    int width, height;
-    glfwGetFramebufferSize(window, &width, &height);
-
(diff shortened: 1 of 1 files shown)


src/example3.cpp
```diff
@@ -81,7 +81,7 @@ int main(int /* argc */, char ** /* argv */) {
 
     int width, height;
     glfwGetFramebufferSize(window, &width, &height);
-    glViewport(0, 0, width, height);
```
wjakob: Probably this should be moved later rather than outright deleting it?

lmurmann: It is called on every frame in Screen::drawWidgets. The glViewport call in the initialization code seems unneccessary. Actually, the two lines above
```
int width, height;
glfwGetFramebufferSize(window, &width, &height);
```
can be deleted as well, without visible side-effects. These two can of course also be left in to demonstrate the API usage.

wjakob: This is incorrect: the point of these lines is to clear the screen once before the mainloop. Application startup often takes a second or so (e.g. to load textures and other resources), and annoying garbage can be seen in the meantime on some platforms without this step.

lmurmann:
Ah, interesting about clearing before entering the main loop. Although I don't see how the three lines above contribute to a framebuffer clear.

Wouldn't you just clear the framebuffer with 
```
glClearColor() -> glClear() -> glfwSwapBuffers() ?
```

I added this sequence right after `Screen::initialize`.

wjakob:
should be fixed by #141. Ignore what I said about clearing the screen.

PR #129, opened 2016-09-16
Small improvements: the missing text scissors and border control on widgets

VirgiliuC:
Hi there,

I have added some small improvements, as described below (see images):

- Added missing scissors on the text labels for 'Button' (and its subsequent descendants, like ComboBox) and for 'CheckBox'. It's better than having text spilling out all over the neighbours.

- Added a 'Show Border' feature for 'Widget' as a base class. The rendering functionality was already there, but wrapped into pre-processor directives, so it was like either compile everything with a (red) border or don't. I believe, in some circumstances, it's good to have a 'controlled' border appearance on some widgets, hence the border on/off and border colour setting via Widget's API. It's a very small addition, that could be useful.
See before/after captures below:

Before:
widget_borders_before

After (window grey border, label cyan border and text clipping):
widget_borders


VirgiliuC:
Some errors in py_doc.h (which I don't use, so i couldn't spot them). To be fixed

PR #133, opened 2016-09-17
Use Color for Screen::mBackground

### LINKED ISSUE #132: Screen background in Python
`Screen::background` returns `Vector3f`, which is not exposed to Python. So neither `background`, nor `setBackground` are usable from Python. Why not use `Color` in those functions?

mrzv:
Fixes #132

---
This change is [Reviewable](https://reviewable.io/reviews/wjakob/nanogui/133)


wjakob:
Great, thank you.

PR #134, opened 2016-09-17
Add constructors from list and tuple for Vector2{i,f} + implicit conversions for all Vectors

### LINKED ISSUE #131: Python bindings for Vector{2i,2f,Xf}
I'm tempted to add constructors to the `Vector`s that take `std::tuple` or `std::vector` (the latter in the case of `VectorXf`). Then I'd like to add implicit conversions, like `py::implicitly_convertible<std::tuple<int,int>,Vector2i>()`. Before I do this, I wanted to check if you find such additions reasonable. The main motivation is that I get tired of wrapping values in `Vector2i` when passing them to various functions.

On a tangentially related note, should the constructor for `Vector2f` be `py::init<float,float>`, rather than `py::init<int,int>`?

mrzv:
Fixes #131

### DIFF
--- a/python/example1.py
+++ b/python/example1.py
@@ -15,10 +15,10 @@ import time
 import gc
 
 from nanogui import Color, Screen, Window, GroupLayout, BoxLayout, \
-                    ToolButton, Vector2i, Label, Button, Widget, \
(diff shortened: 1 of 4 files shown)


wjakob:
LGTM

PR #139, opened 2016-09-25
minor cmake improvements

michalkaptur:
I have added 2 quite useful improvements to nanogui build system:
- cmake uses ccache if only it's availble. It speeds up compilation time a lot.
- all external headers are included with SYSTEM directive, which replaces "-I" with "-isystem". gcc and clang treats these headers as external ones and suspends all compilation warnings (as they are read-only from project point of view).


wjakob:
@michalkaptur: I'm not a big fan of the -isystem change since it only affects include files while "CMAKE_C_FLAGS" is also propagated to the subproject and used during the main compilation (i.e. warnings in .c or .m files). The CCache change is reasonable, but please don't print output when ccache is not used -- this project is used in many other projects with recursive cmake builds, and I'd like to keep the default build as quiet as possible.

michalkaptur:
@wjakob I have removed -isystem to your liking, but still - in my opinion it's a useful habit to include external headers that way.
I also see your point about build verbosity, so I've removed messages from ccache configuration.

wjakob:
Merged, thanks.

wjakob:
This turned out to cause countless warnings in some of my higher-level projects, so I've had to revert this PR.

michalkaptur:
@wjakob Could you provide some error logs or configurations in which this changes cause problems?

wjakob:
I don't have them anymore, but the issue is mentioned in the commit message (and seems to be a well known problem for ccache). Compiling code that is already preprocessed with -Wall -Wextra will sometimes issue warnings that would not have appeared when compiling the original code.

michalkaptur:
Oh, I see. Thank for the response.

PR #141, opened 2016-09-28
Initialize glad before glViewport (fixes #136)

### LINKED ISSUE #136: crashed for example3 in line "    glfwSwapInterval(0);"
It seems 0000000 in Call Stack.

bsdbeard:
When NANOGUI_GLAD is defined, example3 will crash unless:

1. gladLoadGLLoader((GLADloadproc) glfwGetProcAddress) is called before any other OpenGL calls
1. nanogui::Screen constructor finishes before any other OpenGL calls


wjakob:
great, thank you.

PR #144, opened 2016-10-03
Added ColorList for easy access to Colors value

dalerank:


wjakob:
Hi @dalerank,

I've had a closer look at your PR now and don't feel that it's really something that NanoGUI needs to have (it's "nano" after all, so there is an intended lack of features compared to other GUI toolkits).

Feel free to fork the repository to keep it in your own copy though.

Best,
Wenzel

PR #147, opened 2016-10-15
fix missing library on BSD systems

tony:
On FreeBSD (haven't tried Linux yet) this library will be missing. I believe it's only available on OS X.

http://libcxxabi.llvm.org/


tony:
Builds work in Linux, so this seems to be freebsd only.

I'm going to also test this on openbsd as well for good measure.

tony:
OpenBSD 6.0

screen shot 2016-10-22 at 1 41 00 am

Let me brush up tomorrow and test across OpenBSD 6.0, FreeBSD 11-STABLE, MacOS and Ubuntu 16.10

wjakob:
Hi @tony,

this looks great. One note: I've found that "-lc++abi" is needed on Linux (when using Clang), and your patch removes it for that platform.

Let me know when this is ready to merge.

Best,
Wenzel

tony:
Hi @wjakob. I've tested on OpenBSD 6.0, FreeBSD 11-CURRENT, OS X, however have had issues with testing on Ubuntu due to lack of OpenGL 3.3 support in VirtualBox.

wjakob:
Hi @tony,

this looks great now. No worries about Ubuntu, I am using that frequently myself and everything works fine.

One quick question after looking at the PR once more: I saw that you actually completely disabled ``libc++`` when using Clang as a compiler on *BSD. Is that really the right thing to do? To my knowledge, when using Clang, you'll generally want to combine it with libc++ that was developed specifically to be used with that compiler.

I've run into many incompatibilities in the alternative libstdc++ / Clang combination on Linux, hence this setting in the CMake script.

Best,
Wenzel

wjakob:
(note that this comment is unrelated to the `` -lc++abi`` flag which indeed does not seem to be supported on *BSD)

tony:
FreeBSD 11-STABLE will say ``cc: warning: argument unused during compilation: '-stdlib=libc++'`` as a warning. https://lists.freebsd.org/pipermail/freebsd-bugs/2013-October/054196.html

So I can add it back, but we'll get the warning.

>> I saw that you actually completely disabled libc++ when using Clang as a compiler on *BSD. Is that really the right thing to do? To my knowledge, when using Clang, you'll generally want to combine it with libc++ that was developed specifically to be used with that compiler.

I'm not 100% sure. I omit including ``libc++`` by hand on OpenBSD and FreeBSD (as in my current patch), then how is it building and running on FreeBSD and OpenBSD?

I know CMake does *some* magic when it comes to adding flags on your behalf. I can't think of a time I've had the need to go in and add the standard lib stuff manually.

[libc++abi-build.txt](https://github.com/wjakob/nanogui/files/549268/libc.abi-build.txt) is my ``cmake --system-information`` output.

wjakob:
Is this OpenBSD ``cc`` actually Clang?

tony:
Default compiler on OpenBSD is going to be GCC 4.2... So yeah that's a caveat. My testing requires installing llvm (which includes clang).

This may change in OpenBSD 6.1 (https://www.phoronix.com/scan.php?page=news_item&px=OpenBSD-Adds-LLVM-To-Base)

On OpenBSD I am using ``CC=/usr/local/bin/clang CXX=/usr/local/bin/clang++`` with python off.  I was only really intending preliminary OpenBSD support in this patch. I may have promised too much :smile:

tony:
A few things, 

1. current master is broken with FreeBSD. So this PR remedies that.
2. I'm currently in the process of going through OpenBSD setup and see what's needed. *Default* compiler isn't going to support C++11. It's common to install gcc5 and llvm (includes clang) after install due to these limitations.
3. I am going to write docs for installation on OpenBSD and FreeBSD and QA all systems once again.

CMakeLists.txt
```diff
@@ -175,10 +175,14 @@ if (NANOGUI_BUILD_SHARED)
 endif()
 
 # Always use libc++ on Clang
-if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+if (CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND NOT ${CMAKE_SYSTEM_NAME} MATCHES "BSD")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++")
```
tony: On OpenBSD and FreeBSD, this will raise a warning ``clang-3.8: warning: argument unused during compilation: '-stdlib=libc++'``

CMakeLists.txt
```diff
@@ -175,10 +175,14 @@ if (NANOGUI_BUILD_SHARED)
...
-  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -stdlib=libc++ -lc++abi")
-  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -stdlib=libc++ -lc++abi")
+  set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -stdlib=libc++")
+  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -stdlib=libc++")
```
tony: Even when removing these, OpenBSD 6.0 will do this on linking:

```bash
Scanning dependencies of target nanogui
[ 88%] Linking CXX shared library libnanogui.so
/usr/bin/ld: cannot find -lc++
clang-3.8: error: linker command failed with exit code 1 (use -v to see invocation)
*** Error 1 in . (CMakeFiles/nanogui.dir/build.make:157 'libnanogui.so')
*** Error 1 in . (CMakeFiles/Makefile2:107 'CMakeFiles/nanogui.dir/all')
*** Error 1 in /root/study/c++/nanogui/build (Makefile:128 'all')
```

tony:
- Added docs
- Tested again on FreeBSD 11-STABLE, OpenBSD (GCC 4.7 and Clang 3.8) and MacOS Sierra
- Did some more research, the ``-stdlib=libc++`` flag won't work on the BSD's. It will fail compilation phase. I think for the BSD's letting clang find the standard lib is best choice.

wjakob:
Merged, thank you!

PR #151, opened 2016-10-19
Aux drawing function

fairlight1337:
This removes the need to call `Widget::draw()` from `Widget` subclasses by adding an auxiliary drawing function that does the housekeeping for drawing.

I tried to adhere to your coding style as closely as possible.

### DIFF
--- a/include/nanogui/popup.h
+++ b/include/nanogui/popup.h
@@ -53,6 +53,7 @@ public:
 
     virtual void save(Serializer &s) const override;
     virtual bool load(Serializer &s) override;
+    virtual bool drawsFirst() const override { return true; }
 protected:
     /// Internal helper function to maintain nested window position values
     virtual void refreshRelativePlacement() override;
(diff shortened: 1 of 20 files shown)


wjakob:
Hi @fairlight1337 ,

can you motivate this patch a bit? It wasn't immediately obvious to me why you made this change.

Thanks,
Wenzel

fairlight1337:
Hey Wenzel,

Sure, I'll try. Reading the guidelines of what nanogui wants to achieve, I assumed that using less code in subclasses makes it more accessible for contributors. Also (as I keep forgetting these things easily myself) remembering to draw the superclass's overridden function can result in weird bugs that are not always easy to trace.

That was my main motivation. Or is this making the drawing code too complicated?

Jan

wjakob:
Hi,

thanks for the explanation. In this case, I think I'd prefer the old behavior. Although your patch simplifies the default implementation of widgets, it adds some API complexity by being different from how the drawing logic is implemented in other libraries (e.g. Qt, Java/Swing).

Best,
Wenzel

fairlight1337:
Hi,

Thanks for the remarks. I see that point, you're probably right that the API complexity outweighs the gains in widget simplicity.

Cheers,
Jan

PR #153, opened 2016-10-21
GLCanvas widget

fairlight1337:
Added a widget for displaying an OpenGL scene within in, separately from the main GLFW window.

selection_081

If there is anything crucial missing, I'm happy to add it. Serialization should work and with time, I'll add a couple more options (like drawing the border or not and setting callback functions for click, drag, release, etc.).

The widget itself right now just allows a callback (lambda) to be defined where the actual GL drawing code resides. No configuration effort beyond that, except maybe the background color.

Resizing does not yet change the GL texture; I didn't find any best practice in the other widgets on how to do that yet. This is a must though and I'll add it later on.

I also added `example4` which just shows this widget in action. Could be added to one of the other examples as well instead, but they seemed a bit crowded already.


fairlight1337:
There are still a bit of cleanup to be done, in the sense of freeing buffers in the destructor and the like.

I'll add that later this weekend and also look after the points mentioned above.

fairlight1337:
I'm retracting the PR until the widget is actually done. I realized there's just too much still missing to be worthy of addition. Still, let me know if you have strong objections against the idea in general. Thanks.

fairlight1337:
Okay, everything should be okay now. I'll reopen the PR.

wjakob:
Hi @fairlight1337,

this seems like a great idea. However, a few comments:

1. It seems somehow very complicated to set up a FBO, render to the FBO, and then render a textured quad into the widget (all using manual glViewport,gl* calls). Why not directly render into the space occupied by the widget.

The only reason I can see to use an FBO is if the rendering is somehow decoupled -- it could e.g. be costly to render the GLWidget contents, so that it would only be redrawn occasionally when the underlying data changed.

2. All widgets in this library have corresponding Python bindings, which are missing for this new addition.

3. Indentation is inconsistent (some lines use 2 spaces, some use 4 -- which is correct.)

Best,
Wenzel

fairlight1337:
Hi,

1. Sounds very true, thanks. Using an FBO made sense in my head at the time, but I'll see after a way to get around that.

2. My bad, should have thought of that. I'll add it.

3. I must have missed some; my Emacs uses 2 spaces per default. I'll hunt them down.

Cheers,
Jan

fairlight1337:
Hi @wjakob,

I adjusted everything as discussed (except for th `py_doc.h` just yet; I'll do the `mkdoc.py` thing later), thanks again. I have a problem left, though (and maybe that's just me):
```bash
$ cd build/python
$ python
>>> import nanogui
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: ./nanogui.so: undefined symbol: _ZN7nanogui12VScrollPanel16mouseButtonEventERKN5Eigen6MatrixIiLi2ELi1ELi0ELi2ELi1EEEibi
```
Any idea why that happens? The C++ variant works just fine. I even `sudo make install`ed everything.

Cheers and good night,
Jan

fairlight1337:
So, turns out `py_doc.h` needs to be up to date before this builds. I'll look into this tomorrow, my Python doesn't do what I want right now.

wjakob:
Don't worry about the docs -- you can just leave out the ``D()`` macro calls, and I can add them later.

It would be great if you could also port your ``example4.cpp`` to Python to make sure everything works as expected.

Regarding the ``_ZN7nanogui12VScrollPanel16mouseButtonEventERKN5Eigen6MatrixIiLi2ELi1ELi0ELi2ELi1EEEibi`` error. Is it possible that your Python module is somehow stale? That function doesn't exist anymore (was removed a few commits ago, i.e. it's not even in the API anymore), so it's very strange if the Python module expects it to be there.

fairlight1337:
Indeed, I totally missed this:
```
CMake Warning at CMakeLists.txt:403 (message):
  NanoGUI: not building the Python plugin!
```
So it was stale. Thanks for the pointer.

Trying to compile it brings up new problems: `pybind11` is not finding my Python setup with the current `pybind11` version (checking out `pybind11`/<kbd>master</kbd> finds it, though). I set up a `virtenv` with Python3 where it seems to be happy, but I'm getting odd issues over compiling the module (see https://github.com/wjakob/nanogui/issues/155).

fairlight1337:
Alright, I weeded out everything. What's left now is porting `example4` to Python. I'll do that today night or tomorrow, depends on when I find the time.

Thanks for all the support in this.

fairlight1337:
@wjakob I ported most of the example code to Python. One issue I don't know how to handle is that there is no `glEnable` and the like available right now. At least I didn't find any; is `glutil` hiding this somewhere? I don't want to include `PyOpenGL` just for that and am considering adding the respective calls as functions to the cpp `GLCanvas` (and exposing them to Python).

Do you have any pointers for best practices here?

Thanks

fairlight1337:
Also, any idea why AppVeyor isn't happy anymore all of a sudden? I didn't quite get what it wants to tell me. Travis seems to be OK with the changes. Thanks.

wjakob:
Hi @fairlight1337,

in conjunction with the GLShader helper class, the actual number of remaining openGL calls that are needed to put together a project is very small -- basically things like enabling the depth buffer, alpha blending, etc. as you noticed.

I would suggest that you simply also add bindings for the few functions that you used. There are already bindings for a few constants (GL_TRIANGLES, etc.) at the bottom in glutil.cpp. You can add new constants there and bind other functions like so

``
gl.def("glEnable", &glEnable);
``

That's it! :)

As for the AppVeyor crashes: not sure -- it looks to me like the compiler crashesd. Sometimes this thing is a bit flaky and crashes or times out. I'd try and see if it just recovers when doing the next build.

Best,
Wenzel

fairlight1337:
Hi @wjakob,

Thanks for the great advice. I added the two calls to `glutil.cpp` - Worked like a charm :). I finished porting everything and made the cube a bit nicer to look at. Docs will follow soonish, but functionality-wise its there.

Also, AppVeyor seems to be happy again.

Thanks for your support!

Cheers,
Jan

wjakob:
Great -- the implementation looks nice. I'll wait for the docs before merging, ok?

python/glutil.cpp
```diff
@@ -142,12 +142,19 @@ void register_glutil(py::module &m) {
              D(GLShader, drawIndexed), py::arg("type"),
              py::arg("offset"), py::arg("count"))
         .def("setUniform", &setUniformPy, py::arg("name"),
-             py::arg("value"), py::arg("warn") = true);
+             py::arg("value"), py::arg("warn") = true)
+        .def("glEnable", [](GLShader, py::int_ cap) {
```
wjakob: Actually, it would be nicer to turn these into ``gl.def(..)`` at the bottom. These are really static GL calls that aren't related to GLShader, so they should be in the ``gl`` namespace.

fairlight1337:
Like so?

Alright; I'll probably do the docs tomorrow. Is there any tutorial/README/... on how to use `pybind11`'s `mkdoc.py`?

python/glutil.cpp
```diff
@@ -147,7 +147,15 @@ void register_glutil(py::module &m) {
     #define C(name) gl.attr(#name) = py::int_(GL_##name);
     py::module gl = m.def_submodule("gl");
 
+    gl.def("glEnable", [](py::int_ cap) {
+              glEnable(cap);
+           }, py::arg("cap"));
+    gl.def("glDisable", [](py::int_ cap) {
+              glDisable(cap);
+           }, py::arg("cap"));
```
fairlight1337: @wjakob Okay, I put them here. The example is adapted as well and works nicely.

wjakob: Yep, that looks good. You can also replace ``py::int_`` by ``GLenum`` in these two definitions, which is the right type in this case.

fairlight1337: Ah awesome, that looks so much better :)

fairlight1337:
Okay, now all that's missing *should* be the `mkdoc.py` documentation. No idea about that one. Also, wenn I call it like this (python2/3, no idea if that's correct):
```bash
python ./../ext/pybind11/tools/mkdoc.py ../include/nanogui/glcanvas.h
```
It spits all sorts of odd stuff at me:

python2.7
```python
  File "./../ext/pybind11/tools/mkdoc.py", line 230
    print('Processing "%s" ..' % self.filename, file=sys.stderr)
                                                    ^
SyntaxError: invalid syntax
```

python3.4
```python
Traceback (most recent call last):
  File "./../ext/pybind11/tools/mkdoc.py", line 14, in <module>
    from clang import cindex
  File "/usr/local/lib/python2.7/dist-packages/clang/cindex.py", line 515
    raise ValueError,'{0} value {1} already loaded'.format(
                    ^
SyntaxError: invalid syntax
```

fairlight1337:
This keeps growing, I apologize for the large-ish pull request.

wjakob:
No problem about the PR size-- this is obviously a very useful feature.

wjakob:
(see the separate comment about Eigen 4x4 matrices)

fairlight1337:
Okay, I give up on `mkdoc.py`.
```python
Traceback (most recent call last):
  File "./../ext/pybind11/tools/mkdoc.py", line 14, in <module>
    from clang import cindex
  File "~/Workspace/nanogui/python/.venv/lib/python3.4/site-packages/clang/cindex.py", line 515
    raise ValueError,'{0} value {1} already loaded'.format(
                    ^
SyntaxError: invalid syntax
```
Python3.4 is properly set up here. Is that something I can ask you to run as you know what it needs (and probably have a working setup for that as well)? Thanks already.

wjakob:
Don't worry about ``mkdoc.py``. It is very fragile, I'll run it here.

fairlight1337:
Great, thanks. Anything else I should do for the PR? I think I covered everything.

wjakob:
Hi @fairlight1337,

I had one last look before merging and discovered a slight discrepancy. Sorry for not bringing this up until now:

NanoGUI uses callbacks only for semantic events (like "combo box item selected", etc.). In Qt, the corresponding concept is a Signal/Socket pair. In Java/Swing, the corresponding concept is an ActionListener, etc.

Lower level functionality like mouse movement or clicks (which are still uninterpreted, i.e. non-semantic) is handled by virtual functions. The way to customize these in a class is by subclassing the widget.

If that design principle is carried out in your codebase, it will mean that stuff like ``void setMouseButtonCallback(std::function<void(const Vector2i&, int, bool, int)> fncMouseButtonCallback);`` needs to go. In the example4 (both Python and C++ versions), simply subclass GLWidget and add a custom ``mouseButtonEvent`` function. Events will automatically be delivered there thanks to virtual function dispatch & pybind11 magic ;).

It's similar for the ``draw`` function. Instead of a dedicated callback, the way to go is to introduce a new ``virtual void drawGL();`` function with an empty default implementation.

Finally, one more thing: the top-level comments in ``glwidget.h`` are what is used to build the documentation. Currently it says "Canvas widget for rendering GL", which seems a bit on the brief side :). You could say 1-2 sentences about how one can hook up such a widget with actual drawing code.

Best,
Wenzel

wjakob:
To get the new OpenGL draw calls forwarded to Python, you'll need to replace the following line
```cpp
DECLARE_WIDGET(GLCanvas);
```

with
```cpp

class PyGLCanvas : public GLCanvas {
public:
    using GLCanvas::GLCanvas;
    NANOGUI_WIDGET_OVERLOADS(GLCanvas);

    bool drawGL() {
        PYBIND11_OVERLOAD(bool, GLCanvas, drawGL);
    }
};
```

fairlight1337:
Hi @wjakob,

Okay, I see. That makes sense - I'll adjust the code accordingly, probably tomorrow. Thanks for the elaborate guide on what is to be done :).

I'll keep you posted about the progress.

Cheers,
Jan

fairlight1337:
Great, thanks. Anything else I should do for the PR? I think I covered everything.

fairlight1337:
Okay, I updated the `cpp` part. The Python stuff follows :).

fairlight1337:
Hi @wjakob,

Okay, everything is up and running again with subclassing rather than registering callbacks. Both examples (`cpp` and `py`) work like a charm. I added more information in the comments about `GLCanvas` and added doc strings where necessary.

The mouse event related code is gone completely for now, as mouse interaction is not really part of this PR. I'll add the necessary code to all elements in the Arcball example once I get that finished up.

Cheers,
Jan

wjakob:
Beautiful -- thanks for making those changes, it's a big improvement. Do you want me to merge this now or wait for the Arcball change?

fairlight1337:
Thanks for the great pointers on what to change to make it fit.

I'd make the Arcball thing a separate PR to keep things tractable. Also, this PR is self-contained enough I guess.

wjakob:
Got it -- merged!

fairlight1337:
Thanks for the great pointers on what to change to make it fit.

I'd make the Arcball thing a separate PR to keep things tractable. Also, this PR is self-contained enough I guess.

PR #154, opened 2016-10-23
`else()` and `endif()` upside down when deciding whether to build python plugin

fairlight1337:
This just mixed up whether to display the "NanoGUI: building the Python plugin." message or not (it would always show), but still.


wjakob:
...right! Thanks :)

PR #158, opened 2016-10-25
bin2c needs absolute path

devmiyax:
My environment( my be others ) has bin2c already. so I think the abolute path is needed for bin2c in CMakeLists.txt.

```
$ which bin2c
/usr/local/cuda-6.5/bin/bin2c
```


wjakob:
Good catch. You'll need to use a generator expression for this to work on windows, i.e. ``$<TARGET_FILE:bin2c>``

wjakob:
Looks good, thanks -- merged.

PR #159, opened 2016-10-27
Scissor children

fairlight1337:
When forcing a fixed size onto a widget, its child widgets can still draw outside of its designated area. Using the fix in this PR, the widget restricts that area using `nvgScissor` prior to drawing the children.

Example without/with the fix:

&nbsp;

This example forces a fixed height onto the window titled "Tools".

### DIFF
--- a/src/widget.cpp
+++ b/src/widget.cpp
@@ -209,7 +209,7 @@ void Widget::draw(NVGcontext *ctx) {
             nvgSave(ctx);
             child->draw(ctx);
             nvgRestore(ctx);
-	}
+        }
     nvgRestore(ctx);
 }
 
--- a/src/widget.cpp
+++ b/src/widget.cpp
@@ -201,11 +201,16 @@ void Widget::draw(NVGcontext *ctx) {
     if (mChildren.empty())
         return;
 
+    nvgSave(ctx);
     nvgTranslate(ctx, mPos.x(), mPos.y());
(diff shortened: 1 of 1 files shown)


fairlight1337:
The extra pair of `nvgSave`/`nvgRestore` for each child was inserted as the child widgets now also manipulate the scissor region. The initial `nvgTranslate` is undone by the final `nvgRestore` that also restores the scissor region.

wjakob:
I think there are some other widgets that use nvgScissor. Can these statements now be removed?

fairlight1337:
I haven't seen those yet, but I'll gladly go through the source tonight and weed them out.

fairlight1337:
With proper validation of the approach along the way of course ;).

fairlight1337:
Okay, I checked and found that the only other widget using this is `ImageView`. But since it clears the NanoVG drawing stack [right at the beginning](https://github.com/wjakob/nanogui/blob/master/src/imageview.cpp#L284) of the `draw()` method, the former `nvgScissor` call has no effect. It needs its own scissoring to limit the area

 * Where the image is rendered and
 * Where the border is drawn.

I tried removing either, only to find that then there is no scissoring happening at all (probably due to the reason mentioned above).

So I would say there's not much to remove in terms of other scissor calls.

wjakob:
Great, thank you!

PR #160, opened 2016-10-28
[docs] adding missing makefile from original PR https://github.com/wj…

svenevs:
…akob/nanogui/pull/124

Added a few things to the gitignore that don't need to be tracked in the docs folder as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,8 @@ nanogui*.pyd
 libnanogui.so
 libnanogui.dylib
 .vscode/
+# html documentation specific
+*.pyc
+docs/_build
+docs/doxyoutput
+docs/generated_api
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,0 +1,225 @@
+# Makefile for Sphinx documentation
+#
+
+# You can set these variables from the command line.
+SPHINXOPTS    =
+SPHINXBUILD   = sphinx-build
+PAPER         =
+BUILDDIR      = _build
+
+# Internal variables.
+PAPEROPT_a4     = -D latex_paper_size=a4
+PAPEROPT_letter = -D latex_paper_size=letter
+ALLSPHINXOPTS   = -d $(BUILDDIR)/doctrees $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
+# the i18n builder cannot share the environment and doctrees with the others
+I18NSPHINXOPTS  = $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) .
+
+.PHONY: help
+help:
+	@echo "Please use \`make <target>' where <target> is one of"
+	@echo "  html       to make standalone HTML files"
+	@echo "  dirhtml    to make HTML files named index.html in directories"
+	@echo "  singlehtml to make a single large HTML file"
+	@echo "  pickle     to make pickle files"
+	@echo "  json       to make JSON files"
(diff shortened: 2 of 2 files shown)


wjakob:
Merged. However, I was not able to get all the way to HTML output.

After running ``make html``, I get:

```
Generating graph info page...
Generating directory documentation...
Generating index page...
Generating page index...
Generating module index...
Generating namespace index...
Generating namespace member index...
Generating annotated compound index...
Generating alphabetical compound index...
Generating hierarchical class index...
Generating member index...
Generating file index...
Generating file member index...
Generating example index...
finalizing index lists...
lookup cache used 51/65536 hits=102 misses=52
finished...
```

```
$ ls -la _build/html
total 0
drwxr-xr-x 2 wenzel staff  68 Oct 30 11:16 .
drwxr-xr-x 3 wenzel staff 102 Oct 30 11:16 ..
```

It does create a ``html`` directory directly within ``docs``, but that seems to be from doxygen.

svenevs:
I think breathe or exhale is silently failing. At some undetermined point `make html` triggers the `setup(app)` at the bottom of `conf.py`. it's documented somewhere with sphinx-build...

At that point:

1. `doxygen` is generated into `doxyoutput`
    - the output above is from doxygen.
2. Exhale imported and called which in turn is importing breathe
    - If you try-except all of `setup`  and post a traceback I might be able to figure it out

Are you using Python 3? If so try installing sphinx-build for Python 2. Or try importing breathe and exhale in the interpreter (for 2 and 3)? I only ever got it to work for Python 2 :/

During the development I remember the sphinx-build failing silently in strange cases.

svenevs:
To clarify, the `_build/html` folder is created by Sphinx. If you don't have a folder next to `_build` and `doxyoutput` called `generated _api`, then exhale never created anything. Since something in `setup(app)` went catastrophically wrong, sphinx-build halted and i don't think it makes any html?

I believe importing breathe is the issue, but am not sure. Seems my mini library is less portable / error proof than I'd hoped... Let me know if any of the above makes progress and I'll update exhale to be more resilient

wjakob:
Ok, found the issue: it works with Python 2.7 but not 3.4. That's is somewhat problematic -- 2.7 was released over a decade ago and is of course deprecated/discouraged. RTD might switch to 3.x at some point in the future. It would be great if you could take a look to see what's causing the setup to fail with Python 3.

Thanks,
Wenzel

svenevs:
Yes, I agree.  Breathe isn't compatible with python 3, but I don't think it was major stuff.  It was using the now no longer possible way of iterating k,v pairs in dictionaries or something.  I know that per module you can use nice packages such as `six` to avoid this and/or iterating and whatnot.  It isn't possible to statically redefine things in other modules though is it?

I'm pretty sure I was careful to write exhale for python 2 and 3, but obviously never really tested it.  I can take a look at breathe again at some point in the future and see how painful it would be, but the framework was originally autogenerated or something and has a lot of confusing indirections.

wjakob:
Hmm, I see. I am worried that this will just break in the long run -- Python 2 is deprecated and going away for good eventually -- I'm actually hoping that this will happen sooner rather than later.

svenevs:
Valid concerns, and I agree -- py2 is deprecated.

I'll revisit the original attempt at making breathe python3 compatible later this week......but I don't know how bloody that could get.  I don't think it'll be deprecated on RTD in the immediate future, so if anything I can spend more time squaring that off between this semester and next.

PR #167, opened 2016-11-05
Update glad

Snektron:
This update seems to fix compile errors on Windows
Compiler used: MSYS2 GCC 5.3.0


wjakob:
This actually removes some of the modifications to glad that we made for it to work on various versions of MSVC

wjakob:
(Also, a general comment: a lot of the changed lines are just noise -- changes in function name formatting. It would be good if you could crunch it down to the part that is actually needed for GLFW compatibility)

wjakob:
I'll close this for now. Feel free to comment here if you have provided an updated version and I'll reopen the PR.

PR #169, opened 2016-11-08
[docs] updating documentation system for python3

svenevs:
This commit fixes some small oversights that prevented the documentation
from building with python3, remedying concerns in https://github.com/wjakob/nanogui/pull/160

This commit is tested and working here: fake-nanogui.rtfd.io

Further debugging of exhale has revealed why certain functions / typedefs / enums etc will
link to their defining file locally, but not on RTD.  Cause is clear, solution is not.
Will have to consult the folks at RTD to see if they have any input on how to get doxygen
to behave 'normally'.

Full summary:
- Adding better clean target to the Makefile
- Fixed generateDoxygenXML in conf.py
    - This was actually the problem all along, breathe was python3 compatible
- Update exhale to use StringIO correctly for python 2 and 3
- Setting the stage for full and partial template specialization docs
    - Tested and verified breathe source build via requirements.txt
    - Current version installed same as what is hosted through pip
    - Should be up to date enough for templates, release is ~3 days old
    - Electing to keep 'breathe' rather than git
        - Build time is exceptionally slow, it seems it may rebuild every time?

# Configuring Existing RTD for Python3

- Go to the **Admin** page for nanogui on RTD
    - Advanced Settings Tab
    - Near bottom: **Python interpreter:** select CPython 3.x

Everything else should stay the same, it will automatically reinstall things when you make the switch.

Between the time I started with exhale and now it seems breathe has had some updates, and is fully python 3 compatible :)  By virtue of switching to CPython 3.x through RTD, it will install the newest breathe via `pip3` (I think? If you check the build logs, there are mixed `python3` and `python` calls...)

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,1 +1,2 @@
 breathe
+# git+git://github.com/michaeljones/breathe@v4.3.1#egg=breathe
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -46,7 +46,7 @@ help:
 
 .PHONY: clean
 clean:
-	rm -rf $(BUILDDIR)/*
+	rm -rf $(BUILDDIR)/ doxyoutput/ generated_api/ __pycache__/ *.pyc
 
 .PHONY: html
 html:
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -318,8 +318,8 @@ def generateDoxygenXML(stripPath):
     from subprocess import PIPE, Popen
     try:
         doxygen_cmd = ["doxygen", "-"]# "-" tells Doxygen to read configs from stdin
-        doxygen_proc = Popen(doxygen_cmd, stdin=PIPE)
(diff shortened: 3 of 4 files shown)


wjakob:
Dear Stephen,

This is great, thank you so much for porting the docs to Python 3.x. I will switch RTD now.

Best,
Wenzel

docs/requirements.txt
```diff
@@ -1 +1,2 @@
 breathe
+# git+git://github.com/michaeljones/breathe@v4.3.1#egg=breathe
```
svenevs: Yes. Over winter break I plan to get the templates working and wanted to leave that there in case I forgot how to do it or if somebody else wanted to give it a go. But it can definitely be deleted, I doubt I'll forget...

Testing seemed to indicate that it was reinstalled every time the docs were built so it should probably just be removed forever

svenevs:
Sure thing, Python 3 FTW!!!!!

PR #171, opened 2016-11-09
Fix CMake config script to include NanoVG during Python link stage.

milkowski:
It seems there is missing object file nanovg.c.obj when Pythong bindings ale linked. This fix solves the problem. Tested on Win64 with Ninja build tool.


wjakob:
Hi @milkowski,

this is not the right way to do it. NanoVG is already included in the main shared library (though there were some issues related to that).

Can you try again and see if it works with the latest master? If not, can you please post the error messages you got?

Wenzel

PR #173, opened 2016-11-18
Replace bin2c with cmake script

petterarvidsson:
This pull requests replaces bin2c binary with a cmake script that has the same functionality. The rational is mainly for supporting cross compiling.

Pros:
- Less code, less complexity
- Always works as long as cmake itself works

Cons:
- Haven't benchmarked it, but it is for sure a lot slower than the C program (with the small files currently used, not a big issue)
- Slightly less standard solution for generating files with cmake, most people use a binary, except projects with cross compilation needs

For me, this would help a lot with keeping https://github.com/konstructs/client up to date with nanogui. Earlier we used a hack where we built a native binary and then injected it into the path, but since the change to use an absolute path (which I agree is a good change), that doesn't work any more. Therefore I looked at finding a more generic solution that would work in all cases and I do think this one does that.

The PR has been tested on linux and mingw32 (also on linux) cross compilation for windows.  It therefore requires testing with non Makefile based targets, e.g. visual studio, and possibly also Makefile targets on windows and OSX.


wjakob:
Tis is a really neat idea, and I'm super-impressed by your epic cmake skills ;).

petterarvidsson:
@wjakob Glad to hear. There was help coming from an enumerable amount of cmake blogs and stack overflow answers, my contribution was just putting it together for this specific use case.

PR #175, opened 2016-11-30
Preprocessor guards for GL version-specific code

Peaches491:
When attempting to run NanoGUI under GLES2, a number of GL3 functions were not available. This pull guards against those with preprocessor `#ifdef`s, and removes the hard-coded `#define NANOVG_GL3_IMPLEMENTATION` in `src/screen.cpp`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,11 +18,12 @@ else()
   set(NANOGUI_USE_GLAD_DEFAULT OFF)
 endif()
 
-option(NANOGUI_BUILD_EXAMPLE "Build NanoGUI example application?" ON)
-option(NANOGUI_BUILD_SHARED  "Build NanoGUI as a shared library?" ON)
-option(NANOGUI_BUILD_PYTHON  "Build a Python plugin for NanoGUI?" ON)
-option(NANOGUI_USE_GLAD      "Use Glad OpenGL loader library?" ${NANOGUI_USE_GLAD_DEFAULT})
-option(NANOGUI_INSTALL       "Install NanoGUI on `make install`?" ON)
+option(NANOGUI_BUILD_EXAMPLE         "Build NanoGUI example application?" ON)
+option(NANOGUI_BUILD_SHARED          "Build NanoGUI as a shared library?" ON)
+option(NANOGUI_BUILD_PYTHON          "Build a Python plugin for NanoGUI?" ON)
+option(NANOGUI_USE_GLAD              "Use Glad OpenGL loader library?" ${NANOGUI_USE_GLAD_DEFAULT})
+option(NANOGUI_INSTALL               "Install NanoGUI on `make install`?" ON)
+option(NANOGUI_DEFAULT_GL_VERSION    "Default to GL3 compatability?"      ON)
 
 set(NANOGUI_PYTHON_VERSION "" CACHE STRING "Python version to use for compiling the Python plugin")
 
@@ -73,6 +74,11 @@ if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
 endif()
 
 add_subdirectory("${CMAKE_CURRENT_SOURCE_DIR}/ext/glfw" "ext_build/glfw")
+
+# GL Version support: add NANOVG_GL3_IMPLEMENTATION unless disabled
+if (NANOGUI_DEFAULT_GL_VERSION)
+  list(APPEND NANOGUI_EXTRA_DEFS -DNANOVG_GL3_IMPLEMENTATION)
+endif()
 
 # Python support: add NANOGUI_PYTHON flag to all targets
 if (NANOGUI_BUILD_PYTHON)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,12 +18,13 @@ else()
   set(NANOGUI_USE_GLAD_DEFAULT OFF)
 endif()
 
-option(NANOGUI_BUILD_EXAMPLE         "Build NanoGUI example application?" ON)
-option(NANOGUI_BUILD_SHARED          "Build NanoGUI as a shared library?" ON)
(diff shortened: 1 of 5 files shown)


src/screen.cpp
```diff
@@ -43,6 +43,29 @@ std::map<GLFWwindow *, Screen *> __nanogui_screens;
 static bool gladInitialized = false;
 #endif
 
+inline NVGcontext* nvgCreateContext(int flags) {
+#if defined NANOVG_GL2_IMPLEMENTATION
```
wjakob: Can you change this syntax to ``if defined(NANOVG_GL2_IMPLEMENTATION)`` to match the rest of the codebase? (elsewhere as well)

src/screen.cpp
```diff
@@ -35,6 +35,10 @@
 //#define NANOVG_GL3_IMPLEMENTATION
 #include <nanovg_gl.h>
 
+#if defined(NANOVG_GL2_IMPLEMENTATION) || defined(NANOVG_GLES2_IMPLEMENTATION)
+#define NANOVG_CURSOR_DISABLED
```
wjakob: Should be: ``NANOGUI_CURSOR_DISABLED`` since it's our define.

src/screen.cpp
```diff
@@ -32,7 +32,7 @@
 #endif
 
 /* Allow enforcing the GL2 implementation of NanoVG */
-#define NANOVG_GL3_IMPLEMENTATION
+//#define NANOVG_GL3_IMPLEMENTATION
```
wjakob: Better to remove that line completely then.

But I wonder: where is the actual used implementation now defined? (e.g. GL vs GLES, etc.). I assume the CMake build system has to detect and set this somehow..

wjakob:
Hello @Peaches491,

I added a bunch of comments.

Best,
Wenzel

Peaches491:
Hello @wjakob! I believe I've addressed all of you style commits. The only thing left is changes to the CMakeLists file. 

You are correct: The specifiic GL implementation macro is no longer defined. That is likely why the tests are failing. You mention "I assume the CMake build system has to detect and set this somehow." Do you have any idea how that could be done? I could manually specify `NANOVG_GL3_IMPLEMENTATION` like you had before, and that would mirror the current functionality. Users of other GL versions would have to change this themselves. 

What do you think?

Peaches491:
I've added an option to the CMakeList, which, by default, adds `-DNANOVG_GL3_IMPLEMENTATION`. This should mirror the current functionality for existing users.

Peaches491:
@wjakob It appears that the tests are passing now! What do you think about my workaround? Is there an option in CMake which allows the user to specify the GL version? I can only find the boolean `option` tag.

CMakeLists.txt
```diff
@@ -18,11 +18,12 @@ else()
...
-option(NANOGUI_BUILD_SHARED  "Build NanoGUI as a shared library?" ON)
-option(NANOGUI_BUILD_PYTHON  "Build a Python plugin for NanoGUI?" ON)
-option(NANOGUI_USE_GLAD      "Use Glad OpenGL loader library?" ${NANOGUI_USE_GLAD_DEFAULT})
-option(NANOGUI_INSTALL       "Install NanoGUI on `make install`?" ON)
+option(NANOGUI_BUILD_EXAMPLE         "Build NanoGUI example application?" ON)
+option(NANOGUI_BUILD_SHARED          "Build NanoGUI as a shared library?" ON)
+option(NANOGUI_BUILD_PYTHON          "Build a Python plugin for NanoGUI?" ON)
+option(NANOGUI_USE_GLAD              "Use Glad OpenGL loader library?" ${NANOGUI_USE_GLAD_DEFAULT})
+option(NANOGUI_INSTALL               "Install NanoGUI on `make install`?" ON)
+option(NANOGUI_DEFAULT_GL_VERSION    "Default to GL3 compatability?"      ON)
```
svenevs: Verbiage on option string is slightly confusing, what is being controlled here is nano*vg* version.  Maybe `"Use NanoVG OpenGL 3.x+ definitions"` or something?  The 3+ will work for OpenGL 4, which is why `GL3` is slightly misleading.

Or maybe, at the cost of extreme verbosity, change this to be `NANOGUI_DEFAULT_NANOVG_VERSION` for extra clarity?

svenevs (changes requested):
In terms of getting things automated with CMake, I think it would be more challenging to auto-detect a suitable version then it is worth.  [This project](https://github.com/matus-chochlik/oglplus/blob/develop/config/FindOpenGL.cmake) seems to have a bunch of good stuff you could use to get started, being careful to [adhere to their license](http://www.boost.org/LICENSE_1_0.txt).  It's unclear how stable any of that really is though.

You could just use `set` in CMake ([it looks like this](http://stackoverflow.com/a/8710114/3814202)), but I don't think that always shows up in the gui / ccmake.

Ultimately, since there's only two possible options at this point `option` should be fine?  If it is set to `OFF` then assume `2` is being requested.  It would be good to update the docs on [compilation](https://github.com/wjakob/nanogui/blob/master/docs/compilation.rst#default-configurations), then call it a day.  I doubt nanovg will end up adding GL4 versions etc, it's just the differences between 2 and 3 that are problematic.

Peaches491:
Well shoot. This branch will need some updating from me before it's ready. I accidentally mixed in some other changes into this branch. 

Hold off on merging until further notice. I may just open a separate pull

wjakob:
Hi @Peaches491,

any updates here? Also, two requests: could you remove the stb_image* changes? That approach really doesn't work well for shared libraries and will need some modifications of those header files (declspec(dllexport), etc..)

As for the OpenGL standard of choice: let's just set it to GL3 by default, and then users can specify GLES if really needed.

What is the zooxco nanovg fork? Probably better to do that in a separate PR as well?

Best,
Wenzel

PR #183, opened 2016-12-07
Add resize callback to Screen

### LINKED ISSUE #182: Add resize callback to Screen
It seems to me that `Screen` needs a resize callback. Right now the only way I see to handle that event is to subclass `Screen`, which seems like an overkill for such a simple operation. (To be clear, I'm volunteering to do this; I'm starting this issue to make sure I'm not missing anything.)

mrzv:
Fixes #182. The only missing bit are the changes to `python/py_doc.h` (and corresponding `D(...)` bit in `python/widgets.cpp`) because I can't get `doc/mkdoc_rst.py` to work.

---
This change is [Reviewable](https://reviewable.io/reviews/wjakob/nanogui/183)

### DIFF
--- a/include/nanogui/screen.h
+++ b/include/nanogui/screen.h
@@ -123,7 +123,7 @@ public:
     virtual bool keyboardCharacterEvent(unsigned int codepoint);
 
     /// Window resize event handler
-    virtual bool resizeEvent(const Vector2i& size) { if (mResizeCallback) mResizeCallback(size); return false; }
+    virtual bool resizeEvent(const Vector2i& size);
 
     /// Set the resize callback
     std::function<void(Vector2i)> resizeCallback() const { return mResizeCallback; }
--- a/include/nanogui/screen.h
+++ b/include/nanogui/screen.h
@@ -123,7 +123,11 @@ public:
     virtual bool keyboardCharacterEvent(unsigned int codepoint);
 
     /// Window resize event handler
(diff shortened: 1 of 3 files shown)


include/nanogui/screen.h
```diff
@@ -123,7 +123,11 @@ class NANOGUI_EXPORT Screen : public Widget {
     virtual bool keyboardCharacterEvent(unsigned int codepoint);
 
     /// Window resize event handler
-    virtual bool resizeEvent(const Vector2i &) { return false; }
+    virtual bool resizeEvent(const Vector2i& size) { if (mResizeCallback) mResizeCallback(size); return false; }
```
wjakob: One minor request: can you move the implementation of the virtual function to the ``.cpp`` file?

mrzv: Done, I think.

src/screen.cpp
```diff
@@ -456,6 +456,12 @@ bool Screen::keyboardCharacterEvent(unsigned int codepoint) {
     return false;
 }
 
+bool Screen::resizeEvent(const Vector2i& size) {
+    if (mResizeCallback)
+        mResizeCallback(size);
```
wjakob: Sorry, one more thing: return ``true`` when there is a callback? (as the event was handled)

mrzv: Yup. Done.

PR #184, opened 2016-12-07
Fix Python wrapper typo in "uniform"

longjon:
This looks like a copy-and-paste error: the Python wrappers for both `uniform` and `attrib` do the same thing by wrapping `attrib`; presumably the former should be wrapping `uniform`.


wjakob:
Yes, that's a typo! Good catch, thank you.

PR #174, opened 2016-11-24
Fix windows issues

omochi:
I faced build failures and runtime crashes on windows build.

And I fixed them with this PR. Please review it.

# Problems about Eigen

I faced build failures in example2 and example3.
Error messages are these.

```
Severity	Code	Description	Project	File	Line	Suppression State
Error	C2719	'<_Args_0>': formal parameter with requested alignment of 16 won't be aligned	example3	C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\functional	276	
Error	C2719	'v': formal parameter with requested alignment of 16 won't be aligned	example2	C:\Users\omochi\work\github\wjakob\nanogui\include\nanogui\formhelper.h	183	
Error	C2719	'<_Args_0>': formal parameter with requested alignment of 16 won't be aligned	example2	C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\functional	276	
Error	C2719	'v': formal parameter with requested alignment of 16 won't be aligned	example3	C:\Users\omochi\work\github\wjakob\nanogui\include\nanogui\formhelper.h	183	
```

I faced runtime crash in example1.

In example1, program crashes after launch it immediately with this message.

```
Assertion `(reinterpret_cast<size_t>(array) & (sizemask)) == 0 && "this assertion
is explained here: http://eigen.tuxfamily.org/dox-devel/group__TopicUnalignedArrayAssert.html
     READ THIS WEB PAGE !!! ****"' failed.
```

They are both from this Eigen design. 

Details is explained at this page.

https://eigen.tuxfamily.org/dox/group__TopicUnalignedArrayAssert.html

I fixed them by following workaround in this article.

Commits are these.

- add alignment control for eigen problem 98c4ff3
- fix Arcball pybind type parameter bf99ec1
- change setter to const-ref for eigen types a716aab
- const-ref in pybind  a329718

# Problem about OpenGL

In example3, program crashes after launch it immediately at first OpenGL function calls without any messages.
This is from lack of setting up `GLAD` for windows.

I fixed it in this commit.

- add glad init for windows b00670f

# Problem about build python binding

In my environment, cmake does not detect python library for pybind module.

I solved problem with this.

- fix CMakeLists to build with python for windows bd1e7ef

I don't understand deeply it and cmake. Sorry.
And I could not find out about original value `${MITSUBA_PYTHON_VERSION}`.

---

I checked commits by building and running examples 1 to 4 with my windows PC and Mac.

My environment is below.

## windows

- Windows 10
[...]

PR #185, opened 2016-12-09
Fix retina issue

omochi:
I faced issue that the result of `nvgTextBounds` changes in between `Screen` constructor and `drawWidgets`.
This happens resizing of button if we call `performLayout`.

I investigated the reason of this.

I found that the reason is difference of `devicePixelRatio` in `nanovg`.
This value affect text rasterize result in `nanovg`.

`nvg__setDevicePixelRatio` is private function and only way to call this is call `nvgBeginFrame`.
`nanogui` calls `nvgBeginFrame` in `drawWidgets` first time.
But it defers to send actual `devicePixelRatio` to `nanovg`.
So I added `nvgBeginFrame` and `nvgEndFrame` at the bottom of `Screen` constructor.

This PR also  includes #174.


wjakob:
Merged, sorry -- it took me a while.

PR #186, opened 2016-12-10
Reorder python version list, assume if user has higher python3 versio…

svenevs:
…n they intend to use it.

I primarily operate on Fedora, which by default includes python 3.4.  I recently discovered the [matmul operator](https://docs.python.org/3/library/operator.html#operator.matmul) and decided to upgrade to 3.5.2.

The current build system will grab python 3.4 by default, this switch just makes it so that a higher version takes precedence when building the python bindings.  It seems reasonable to assume that if somebody has a higher version they'd want to use it.

#174 fixes `MITSUBA_PYTHON_VERSION` to be `NANOGUI_PYTHON_VERSION`, and with either available this PR doesn't do too much / may not be warranted...just makes it so I don't have to specify anything explicitly when running `cmake` ;)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -412,7 +412,7 @@ if (NANOGUI_BUILD_PYTHON)
 
   # Try to autodetect Python (can be overridden manually if needed)
   list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/ext/pybind11/tools")
-  set(Python_ADDITIONAL_VERSIONS 3.4 3.5 3.6 3.7)
+  set(Python_ADDITIONAL_VERSIONS 3.7 3.6 3.5 3.4)
   find_package(PythonLibsNew ${MITSUBA_PYTHON_VERSION})
   if (NOT PYTHONLIBS_FOUND)
     # Python not found -- disable the plugin


wjakob:
Merged, thanks!

mrzv:
I have a tangentially related question. Is nanogui supposed to respect `NANOGUI_PYTHON_VERSION`? I see it defined in `CMakeLists.txt`, but I don't see where and how it's used. (It also doesn't seem to affect anything in practice.) In general, if I wanted it to use some older version (e.g., 2.7), what's the correct way to do it? Is it to manually set the three Python variables, or is there some other procedure?

wjakob:
Check again, I just committed a fix related to that.

svenevs:
@mrzv now that the reminant `MITSUBA_PYTHON_VERSION` has been fixed to be `NANOGUI_PYTHON_VERSION`, currently you would have to execute (assuming you are in a `build` directory one level below `CMakeLists.txt`)

```
cmake .. -DNANOGUI_PYTHON_VERSION="2.7.12"
```

And yea as @wjakob just said, [this line](https://github.com/wjakob/nanogui/blob/master/CMakeLists.txt#L27) makes it explicit in the cmake gui's :)

mrzv:
Yup. Works now. Thanks.

PR #194, opened 2016-12-20
[docs] Do not specify bin2c, as it is no longer a target

GreatAttractor:
I am using NanoGUI as a Git submodule in my project, as recommended in `docs/compilation.rst`. The instructions need updating due to the recent removal of `bin2c` as a target.


wjakob:
Thanks!

PR #195, opened 2016-12-22
Fix smart pointer issue

omochi:
I found 2 issues about `nanogui::ref` design.

- 1. const handling
- 2. implicit conversion problem

# const handling

## current design

There are 7 methods for pointer operator overloads.

```
T* operator->();
const T* operator->() const;
T& operator*();
const T& operator*() const;
operator T* ();
T* get();
const T* get() const;
```

These methods design about const is broken.

This prevent pattens such like below.

```
void f1(const ref<Widget> & widget) {
	widget->setEnabled(true);
}
```

This is compile error.
`widget` is const reference, so it can only call const `->` operator.
but it returns `const Widget *`, so you can not call `setWidget`.

This code can compile.

```
void f2(ref<Widget> widget) {
	widget->setEnabled(true);
}
```

But this argument makes unnecessary copy constructor invocation.
Const reference passing is better than value copy passing almost cases in C++.

`const ref<Widget> &` means immutable pointer.
It does not mean immutable pointee object.

We expect that `const ref<Widget> &` means `Widget * const` in raw pointer.
But with this const `->` operator design,
`const ref<Widget &>` means `const Widget * const` in raw pointer unexpectedly.

## proposal design

My proposal is below.

```
T * operator->() const;
T & operator*() const;
operator T* const ();
T * get() const;
```

I delete original 7 methods and create new 4 methods.
These are correct pointer design.

There is a similar solution in C++ std.
`std::shared_ptr<T>` has same design.

If you want const pointer such like `const Widget *`.
You can use `ref<const Widget>`.

# implicit conversion problem

There are 2 implicit conversion of ref.

```
operator T* () const;
operator bool() const;
```

My patch about `const` already applied to first entry.

They causes unexpectedly bug.
Look this code.

```
void f3(const ref<Widget> & parent, const ref<Widget> & child) {
	parent->removeChild(child);
}
```

You can compile this code.
But it may crash.

You desire compiler to call `child.operator Widget * ()` implicitly and select `parent.removeChild(Widget * child)` overload.
But actually compiler calls `child.operator bool()` implicitly, and select `removeChild(int index)`.
Compiler also inserts `bool` to `int` conversion implicitly. So this code means `parent->removeChild(1);`.

## proposal design

I put `explicit` to `bool`.

```
operator T* ();
explicit operator bool() const;
```

With this design, `removeChildren` called as expected.

You can still write this.

```
void f4(const ref<Widget> & x) {
	if (x) {
		printf("not null\n");
	}
}
```

Because `if` condition expression is converted to bool implicitly.

For reference, `std::shared_ptr` also has `explicit operator bool() const`.

# included PR

this PR includes my #193 PR.


omochi:
@wjakob Hello I am working in my previous project with nanogui.
I am refactoring code and facing this issue again.
Please review this PR !

PR #198, opened 2017-01-04
Fix focus issue

omochi:
This PR solves dangling focus issue.

# about dangling focus issue

Screen has `mFocusPath`.
It is not updated when a child widget is removed.
So dangling pointer will be made in `mFocusPath`.
It may cause crash.

# example crash case

Crash case can be made with small modification of example1.

```
class ExampleApplication : public nanogui::Screen {
public:
	Widget * a_;

    ExampleApplication() : nanogui::Screen(Eigen::Vector2i(1024, 768), "NanoGUI Test") {
      	
      	...
        
        auto ib = panel->add<IntBox<int>>();
        ib->setEditable(true);
		a_ = ib;

		...
	}

    virtual bool keyboardEvent(int key, int scancode, int action, int modifiers) {
		if (key == GLFW_KEY_1) {
			if (a_) {
				a_->parent()->removeChild(a_);
				a_ = nullptr;
			}
			return true;
		}

        if (Screen::keyboardEvent(key, scancode, action, modifiers))
            return true;
        if (key == GLFW_KEY_ESCAPE && action == GLFW_PRESS) {
            setVisible(false);
            return true;
        }

        return false;
    }
```

Steps are below.

- click int box to focus.
- push `1` key to run code.

After run `keyboardEvent`, run `keyboardCharacterEvent` immediately and crash in it.

# my fix approach

I check focus at begin of `removeChild`.
If child has focus, move focus to parent.

And, to share this logic in 2 `removeChild` overloads,
I change one to call another one.

# includes

this PR includes #193, #195


FurryFur:
Any reason this hasn't been merged in yet? I just encountered the same issue.

wjakob:
This project is managed by a busy prof who is spread thin amidst dozens of projects, so you'll have to be patient..

wjakob:
Part of the reason is also that this PR does a whole bunch of other things that aren't discussed anywhere.

PR #204, opened 2017-02-10
Allow Screen to receive drag events.

liaody:
This will allow overriding of mouseDragEvent of the Screen object. Very handy for google earth type of application where the background of the whole app is the main display of 3D scene.


liaody:
Sorry the branch is polluted with my other changes. Let me know if this is useful, I can break them out and submit again.

Issue #47, opened 2016-03-02
[Suggestion] Add a step-by-step guide on including it into an existing project.


PatronBernard:
I have tried for hours to start with an empty project and then adding nanogui to it but it did not work. I tried mirroring the configuration properties, but the included examples are so different from a default empty project, so if I want to add it to an existing project I am affraid I will mess up other parts. 

I have no idea what the issue could be or what I should pay attention to. This makes troubleshooting very difficult.  I expected it to be straightforward (adding header and library directories, including nanogui/nanogui.h and linking nanogui.lib) but it is not. Perhaps someone could add a HOWTO on adding it to a VS project to get a minimal working example?

dupred16:
I have to agree, I think I am running into the same problem, it has to do with using glew in an existing project. I have tried static and dynamic linking, and I can get everything to link, I can get nanogui to init fine, but inside my application code none of the glew extensions are initialized. So I tried to make a glfw window first, but as soon as nanogui tries to a use a glew bound call it fails.

PatronBernard:
Yes! My project also

PatronBernard:
..uses GLEW and they seem to conflict.

BennetLeff:
PatronBernard, it's been a while since you opened this issue but if you're still interested I will be working on a way to add some documentation because I just got a build working with nanogui (although not 100% pretty code or seamless with my outside of library OpenGL calls). After I improve my projects code quality considerably I can offer some samples in addition to the provided ones. Does anyone have anything against using something like [sphinx](http://www.sphinx-doc.org/en/stable/#) to at least get the documentation process started?

wjakob:
Using Sphinx sounds good to me.
(BTW: In the meantime, NanoGUI has switched from GLEW to GLAD)

darrenmothersele:
@PatronBernard this is how I got up and running...

    $ mkdir nanogui-test
    $ cd nanogui-test
    $ git init
    $ git submodule add https://github.com/wjakob/nanogui.git lib/nanogui
    $ git submodule update --init --recursive

Then I have a `CMakeLists.txt` file that looks like this:

```
cmake_minimum_required(VERSION 3.5)
project(nanogui_test)

set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")

add_subdirectory(lib/nanogui)
include_directories(lib/nanogui/include)
include_directories(${NANOGUI_EXTRA_INCS})
[...]

PR #206, opened 2017-02-20
[docs] Add link to documentation

mottosso:
Noticed some documentation in the `docs/` folder, and searched the [issue section](https://github.com/wjakob/nanogui/issues/47#issuecomment-244861955) for clues as to where it may be hosted.

Let's make it official!

- https://nanogui.readthedocs.io


wjakob:
There is actually a button, but this may be a little subtle. Adding an extra link sounds like a good idea.

mottosso:
Ah so there is, did not notice it until now. I was also text-searching for "docs" and "readthedocs" but came up empty.

PR #200, opened 2017-01-15
[docs] fixing link_title so templates do not have name trimmed

svenevs:
Hey @wjakob I actually got lucky, not much had to change for things to work...I think.

I don't know if you've looked at the docs recently, but something has gone terribly wrong lately.  I posted [an issue](https://github.com/michaeljones/breathe/issues/303) on breathe's page, but titles for almost everything are getting deleted.

I see two possible options:

1. Hard-code a specific (earlier) version of breathe in `requirements.txt`
2. Try and use [this stuff](https://github.com/arximboldi/immer/commit/e6e3c3f7d8ed4ade23928c3db56ca90108d02b85).  I haven't tested this, and it is definitely not going to fix everything for us here.

I made this a PR to start a discussion, but this should not be merged in as is.

PR #208, opened 2017-02-20
[docs] bad declaration, fixed labels for templates

svenevs:
turned out the labels were the only problem, i think i actually had that in #200 but this is easier to see what is actually changing.

another user found a bad declaration, so that is fixed as well.

not really sure i like this workflow.  tracking `exhale.py` was an easy solution to "where should i actually write files to", but it's kind of a pain for anybody to manage when an update occurs.  do you think it's worth packaging on pypi over the summer when i have more time?

### DIFF
--- a/docs/exhale.py
+++ b/docs/exhale.py
@@ -2,7 +2,7 @@
 #
 # This file was generated on/around (date -Ru):
 #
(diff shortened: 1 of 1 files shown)


wjakob:
Definitely -- installing via PyPi would be the best solution. (and it's pretty easy to set up)

PR #211, opened 2017-04-05
Fixed bug in Framebuffer::downloadTGA()

NSchertler:
There was a cast from some alloca variable that did not exist.


wjakob:
Hi Nico,

``alloca()`` is actually a standard function. See e.g. https://www.gnu.org/software/libc/manual/html_node/Advantages-of-Alloca.html. It's possible that MSVC needs an extra header file to support it (and that fix would be preferable to the one you've posted now).

Best,
Wenzel

NSchertler:
Ah, I see. Was compiling with MinGW and assumed that MSVC didn’t complain because the function wasn’t needed.

PR #216, opened 2017-05-02
mainLoop refresh timeout 

xSaiteix:
Added glfwWaitEventsTimeout call. 

If you pass refresh value <= 0 then update loop runs on its maximum rate. 
In some cases there is no need to wait for keyboard/mouse/etc. events.
For example: custom OpenGL widget that draws animated scene.   

glfwWaitEventsTimeout(0) == glfwPollEvents();

### DIFF
--- a/src/common.cpp
+++ b/src/common.cpp
@@ -20,6 +20,7 @@
 #include <thread>
 #include <chrono>
 #include <iostream>
+#include <algorithm>
 
 #if !defined(_WIN32)
     #include <locale.h>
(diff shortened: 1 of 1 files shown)


xSaiteix:
Why AppVeyor throws errors?:
```
C:\projects\nanogui\src\common.cpp(93): error C2589: '(': illegal token on right side of '::' [C:\projects\nanogui\nanogui-obj.vcxproj]
C:\projects\nanogui\src\common.cpp(93): error C2059: syntax error: '::' [C:\projects\nanogui\nanogui-obj.vcxproj]
```

Line 93:
`const double timeout = std::max(refresh, 0) / 1000.0;`

PR #217, opened 2017-05-03
Update to Eigen 3.3.3

haldean:
This updates nanogui to work with Eigen 3.3.3. The only code change required was a small bit in screen.cpp. 
 Eigen 3.3.3 doesn't support multiplying or dividing an integer vector through by a float; this fixes the code that relied on that behavior by first casting the `mSize` vector to a float type, multiplying by the pixel ratio, and then casting that result back to an integer.


wjakob:
Sorry for the big delay. The Eigen version was updated in the meantime.

PR #220, opened 2017-05-15
fixes for compiling with cmake generated Xcode project

AndySomogyi:
Hi, 

I added a few fixes that enables compilation of nanogui with CMake generated Xcode projects. Xcode has a bug, and I've implemented the CMake suggested work around. 

I've verified that this fix works for both Xcode generated projects, and CMake generated makefiles.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -336,10 +336,27 @@ add_library(nanogui-obj OBJECT
   src/serializer.cpp
 )
 
-add_library(nanogui ${NANOGUI_LIBRARY_TYPE}
-  $<TARGET_OBJECTS:nanogui-obj>
-  $<TARGET_OBJECTS:glfw_objects>
-)
+# XCode has a serious bug where the XCode project produces an invalid target that will not get
+# linked if it consists only of objects from object libraries, it will not generate any
+# products (executables, libraries). The only work around is to add a dummy source
+# file to the library definition. This is an XCode, not a CMake bug.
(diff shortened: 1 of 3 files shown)


AndySomogyi:
Hi, 

The Travis CI failing test is weird because it looks like the host CMake was possibly downgraded. The nanogui CMakeLists.txt has required version 2.8.12 for some time, but looks like the CMake on the host was changed to 2.8.7. 

Output: 
`
 CMake 2.8.12 or higher is required.  You are running version 2.8.7
`

AndySomogyi:
Hi, 

looks like something strange is going on with the Travis CI, as the nanogui CMakeLists.txt has required CMake version 2.8.12 for some time, but for some reason, the host operating system seems to have been downgraded to version  2.8.7. 

Output from Travis CI on the gcc 4.8 test:
`
CMake 2.8.12 or higher is required.  You are running version 2.8.7
`

wjakob:
Apologies for the big delay. Two issue that I see with the PR are

1. It fails the CI tests (probably easily fixable by rebasing)
2. It hacks the STB implementation exports to across the platforms, rather than switching to a uniform scheme. What would have made much more sense to me is to ensure that the symbols are exported on all platforms, and then import them consistently.

Items #1 and #2 are really independent from each other. For that reason, it may be easier to split this into two separate PRs.

Thanks,
Wenzel

svenevs:
The Travis builds failed because this was submitted before #224. If you close and reopen the PR that will relaunch Travis.

PR #222, opened 2017-06-05
Added Graph::mFill flag to render flag with or without fill.

rvega:
If set to false, the graph plot is rendered with just a line (no fill), the default (true) value, renders the Graph as usual (with fill).


rvega:
Copy pasting from #220:

Hi,

looks like something strange is going on with the Travis CI, as the nanogui CMakeLists.txt has required CMake version 2.8.12 for some time, but for some reason, the host operating system seems to have been downgraded to version 2.8.7.

Output from Travis CI on the gcc 4.8 test:
CMake 2.8.12 or higher is required. You are running version 2.8.7

PR #223, opened 2017-06-05
Added Y offset and scale to Graph

rvega:
By setting Graph::mOffsetY and Graph::mScaleY you can plot vertical ranges different from [0,1]


wjakob:
I understand the motivation here. I don't think it makes sense to bake too much functionality into the core graph widget though -- it will make more sense for you to build your own graph widget if you need scaling/offset functionality (and likely more extra things in the future)

PR #224, opened 2017-06-06
update linux travis build

svenevs:
- Note: this is using `cmake 3.2.2`, not `2.8.12`.
    - It seems it used to come from `kubuntu-backports`, but does not anymore.
    - The only other whitelisted package I could find had `2.8.11`.  See discussion [here](https://github.com/travis-ci/apt-source-whitelist/issues/118)
- It seems you can alternatively install `cmake` from source
    - Meh. It would be nice if they just whitelisted the linked issue.  OSX is using [`cmake 3.6.2`](https://travis-ci.org/wjakob/nanogui/jobs/239823812#L62), so I suppose it is a bad idea not to have `2.8.12` in there.
- No trusted `2.8.12` PPA are available on travis at this time.
- Added some (missing?) linux gfx build deps.

Should allow #220, #222, and #223 to build now.

Build results: [travis-ci on my fork](https://travis-ci.org/svenevs/nanogui/builds/239808315)

**Warning**: see explanation on [travis docs](https://docs.travis-ci.com/user/trusty-ci-environment/), officially it is still in Beta.  I tried it and it just worked TM, but it may not actually be what you want in which case I'll just close this PR.  What about keeping Trusty but building `cmake 2.8.12` from source?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,7 @@ matrix:
     - os: osx
       compiler: clang
       script:
+        - cmake --version
         - cmake -DNANOGUI_PYTHON_VERSION=2.7
         - make -j 2
     - os: linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: cpp
+dist: trusty
 sudo: false
 matrix:
   include:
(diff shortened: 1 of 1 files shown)


wjakob:
Merged -- thank you. Sorry that it took me a while to get to this.

PR #233, opened 2017-07-31
Context menu

austensatterlee:
I recently created a context menu widget for my own project, and I thought it might be beneficial to others. This is just a standard context menu, the type that we are all familiar with. It supports arbitrarily nested submenus, and can delete itself automatically after it is done being used. This means it can be created directly inside an event handler. Here's a screenshot of example1 modified to open a context menu when right clicking the background:
image

The context menu deactivates itself upon losing focus (i.e. when anything other than the menu is clicked on). To get this to work, I needed to make a small modification to the Screen::updateFocus method. The change simply avoids sending a de-focus event to widgets in the old focus path if that widget is also in the new focus path (meaning that widget is not really being de-focused anyway). I personally think this is more aligned with the expected behavior of the focus system, but if there is a reason it shouldn't be done this way I believe I could modify the context menu widget to work with the old behavior.

Unfortunately, I have not included python bindings yet because I'm having trouble with mkdoc.py. Could anyone guide me through the process of how py_doc.h should be generated?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -325,6 +325,7 @@ add_library(nanogui-obj OBJECT
   include/nanogui/stackedwidget.h src/stackedwidget.cpp
   include/nanogui/tabheader.h src/tabheader.cpp
   include/nanogui/tabwidget.h src/tabwidget.cpp
+  include/nanogui/contextmenu.h src/contextmenu.cpp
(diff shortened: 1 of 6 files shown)


austensatterlee:
Hmm, I tried fixing the whitespace issues that travis was previously complaining about, but I'm not sure what is causing it to fail now. Looks like more formatting issues. Can anyone offer guidance?

PR #242, opened 2017-08-02
Window no longer draws shadow if shadow size == 0.

nhojb:
Minor change to avoid calling drop shadow drawing code if ds size is zero.


wjakob:
thanks!

PR #243, opened 2017-08-02
Removing use of -retain (incompatible with ARC).

nhojb:
Non-ARC Objective-C projects are now very rare.

Note: the old code would have leaked without ARC anyway (no autorelease of "types" array).

### DIFF
--- a/src/darwin.mm
+++ b/src/darwin.mm
@@ -6,7 +6,7 @@ NAMESPACE_BEGIN(nanogui)
 std::string file_dialog(const std::vector<std::pair<std::string, std::string>> &filetypes, bool save) {
     std::string path = "";
     if (save) {
-        NSSavePanel *saveDlg = [[NSSavePanel savePanel] retain];
+        NSSavePanel *saveDlg = [NSSavePanel savePanel];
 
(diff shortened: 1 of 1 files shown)


wjakob:
Hi @nhojb,

thank you for all of your contributions. It's going to take me a bit of time to make it through all of these -- for this one, would you mind changing the PR so that it avoids leaks?

Thanks,
Wenzel

nhojb:
Hi Wenzel,

No problem. Hope that the changes are useful.

Regarding this PR, there are no leaks if the file is compiled with ARC enabled (the default). It was only leaking if ARC is disabled - which was necessary because of the use of "retain".

wjakob:
I just checked, and this is not the case at least on my machine (NanoGUI compiled with latest XCode):
```
printf("ARC enabled? : %i\n", __has_feature(objc_arc));
```
prints "0" without an extra ``add_definitions(-fobjc-arc)`` statement in CMake.

wjakob:
(I amended your commit with this CMake flag)

nhojb:
Thanks. Hadn't thought to check CMake. We add nanogui src direct in Xcode (though I have of course been using CMake for these PRs).

PR #245, opened 2017-08-02
Adding check for null Screen which was causing a crash in our project.

nhojb:


wjakob:
Actually, I committed this perhaps a bit prematurely -- do you have any intuition why  ``Screen*`` might be ``nullptr``? It does not appear to be a normal situation in a NanoGUI program.

nhojb:
Not sure. Our use of nanogui is probably unusual in that we use it as part of a plugin for Final Cut Pro X. As such we are creating and tearing down the nanogui stack in response to user actions. It's possible that the invalid state is the cause of this.

Having said that we were encountering this crash frequently and this change did fix the issue. However it may be particular to our usage.

wjakob:
ok, that seems fine given that it is such a tiny change.

wjakob:
Out of curiosity, what's the name of your plugin?

nhojb:
ColorFinale: http://try.colorgradingcentral.com/colorfinale/

We use nanogui for our on-screen-controls. I've been very impressed with the architecture (Qt inspired?) and the quality of the code ;-)

PR #249, opened 2017-08-16
Add additional helper methods for setting shader uniforms

Tom94:
This pull request adds additional helper methods for setting shader uniforms.


wjakob:
Looks good, thank you!

PR #252, opened 2017-08-24
Do not require GLCanvas to have a parent window

Tom94:
Currently, `GLCanvas` needs to obtain the root `Screen` for drawing purposes, but does so by first finding the parent `Window` and then picking the parent of that. This has two problems:

1. A `Window` is not necessarily a direct descendant of `Screen`.
2. There may not be a `Window` between `GLCanvas` and `Screen` (my use case).

The simple fix is to directly recurse up the parents until the screen is found, rather than stopping by the `Window` first.

### DIFF
--- a/include/nanogui/widget.h
+++ b/include/nanogui/widget.h
@@ -162,6 +162,9 @@ public:
     /// Walk up the hierarchy and return the parent window
     Window *window();
 
+    /// Walk up the hierarchy and return the parent screen
+    Screen *screen();
+
     /// Associate this widget with an ID value (optional)
     void setId(const std::string &id) { mId = id; }
     /// Return the ID value associated with this widget, if any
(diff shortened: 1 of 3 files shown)


svenevs:
This looks great!

@fairlight1337 I believe you wrote the original.  This PR doesn't seem to create any issues, is there anything specific that pops out at you?

@wjakob Is it **guaranteed** that (eventually) a `nanogui::Screen` will always be a parent?  I'm concerned about the `while (true)` traversal, or possible a widget somehow having two possible `nanogui::Screen` parents.

If it's always exactly one `nanogui::Screen` and it is always going to be there, this shouldn't be a problem.  If it is always exactly one `nanogui::Screen`, or potentially nothing, the loop can change to be something like

```cpp
Widget *widget = this;
while (widget) {
    Screen *screen = dynamic_cast<Screen *>(widget);
    if (screen)
        return screen;

    widget = widget->parent();
}

return nullptr;
```

This means that anybody calling `Widget::screen()` needs to check if it was `nullptr` or not, though.

fairlight1337:
@svenevs I didn't test it, but can't see any issues there. Looks good indeed!

wjakob:
This looks great -- I can see how a hardcoded ``window()->parent()`` to get the ``Screen`` would cause issues.

PR #255, opened 2017-09-01
Allow accessing the scroll amount of VScrollPanel from outside

Tom94:
What the title says. :)


wjakob:
Oops, I merged this too prematurely. Can you please also add a Python binding for this function? It's a problem when the C++ code and bindings get out of sync.

PR #256, opened 2017-09-03
Allow file_dialog to open multiple files at once and fix undefined behavior with zenity

Tom94:
This PR introduces a little bit of code duplication, but I felt in these specific isolated cases of heavily OS-specific code it is acceptable. If you disagree, I'll be happy to refactor the code a bit. No strong preference.

Cheers!

### DIFF
--- a/include/nanogui/common.h
+++ b/include/nanogui/common.h
@@ -464,6 +464,25 @@ extern NANOGUI_EXPORT std::string
 file_dialog(const std::vector<std::pair<std::string, std::string>> &filetypes,
             bool save);
 
+/**
+ * \brief Open a native file open dialog, which allows multiple selection.
+ *
+ * \param filetypes
+ *     Pairs of permissible formats with descriptions like
+ *     ``("png", "Portable Network Graphics")``.
+ *
+ * \param save
+ *     Set to ``true`` if you would like subsequent file dialogs to open
+ *     at whatever folder they were in when they close this one.
+ *
+ * \param multiple
+ *     Set to ``true`` if you would like to be able to select multiple
+ *     files at once. May not be simultaneously true with \p save.
+ */
+extern NANOGUI_EXPORT std::vector<std::string>
+file_dialog(const std::vector<std::pair<std::string, std::string>> &filetypes,
+            bool save, bool multiple);
+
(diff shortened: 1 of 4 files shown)


Tom94:
I forgot to mention: Tested on Windows, OS X, and Ubuntu.

Tom94:
Re-pushed the last commit to re-run CI. For some reason the previous travis g++ build was set to "Cancelled", which led to CI seemingly failing. Could be related to the recent downtime travis had.

wjakob:
Hmm -- as you suggest, I think that a ``bool multiple = false`` extra parameter would lead to quite a bit less code duplication. I must say I'd much prefer that option. I think should throw an exception if both ``multiple`` and ``save`` are true in that case.

Tom94:
Alright, I merged the implementations of the single and multiple file dialogs. I also kept the old single-parameter function returning a `string` instead of `vector<string>` to not break API backwards compatibility.

At one point I considered using an `enum { OpenSingle, OpenMultiple, Save }` instead of the boolean parameters, which would remove the need for throwing an exception. I decided against it, because it would make the single-file vs multi-file API inconsistent.

Tom94:
I totally forgot about the python bindings. Unfortunately, since python does not permit multiple functions with the same name I reverted to the previous API (`string file_dialog` and `vector<string> file_dialog_multiple`) and moved the 2-boolean shared implementation into an internal function `__nanogui_file_dialog`.

wjakob:
Not sure I get it. Pybind11 allows two functions with the same name (it uses a 2-pass overload resolution scheme).

Tom94:
There is also a good chance I am doing something wrong, as I have not used pybind before.

Could it be a problem that the overloaded functions had different return types? The failed build logs indicate there was a problem with resolving the overloaded function types:
- [GCC](https://travis-ci.org/wjakob/nanogui/jobs/274231980)
- [MSVC](https://ci.appveyor.com/project/wjakob/nanogui/build/1.0.487/messages)

Tom94:
Ah, apologies. I see now how I went wrong. Will revert and fixup.

wjakob:
Looks good, thanks!

PR #260, opened 2017-09-17
Entypo+ and associated defs, many new docs, colorpicker, updated examples

svenevs:
Definitely squash this, there was a lot of intermediate commits for the entypo stuff that have now been reverted or destroyed.

~~**NOTE**: I'm going through and doing a PR review right now.  Don't review until this is gone since this is a huge update and most of it is just noise from doc comments.~~

1. It would be kind of tedious to separate this out into multiple PRs, but it can be done if you prefer.  Things like docs links got added in a few different locations relating to entypo.

2. See annotations in PR review.  Basically, this all boils down to the following:

    - Entypo+ updates
        - Some constants have changed, those are all fixed.  Entypo+ file documentation enumerates everything, as well as there are CPP and python examples.
        - Python binding code changes slightly to allow auto-generation of Entypo+ constants without having to incorporate GLFW stuff in the gen code.
        - **Important**: icon sizes have changed.  See [relevant change to common.h](https://github.com/wjakob/nanogui/pull/260#discussion_r139310396).  Here is a full index of where that applies:
            - [first scale change](https://github.com/wjakob/nanogui/pull/260#discussion_r139310615)
            - [second scale change](https://github.com/wjakob/nanogui/pull/260#discussion_r139310622)
            - [third scale change](https://github.com/wjakob/nanogui/pull/260#discussion_r139310634)
            - [fourth scale change](https://github.com/wjakob/nanogui/pull/260#discussion_r139310724)
            - [fifth scale change](https://github.com/wjakob/nanogui/pull/260#discussion_r139310729)
            - [sixth scale change](https://github.com/wjakob/nanogui/pull/260#discussion_r139310735)
            - [seventh scale change](https://github.com/wjakob/nanogui/pull/260#discussion_r139310748)
            - Basically, the ones using `ih` should be good.  But a couple were not scaled the same as previous code.
    - ColorPicker has been modified
        - I believe the new functionality makes sense.  These changes came about from writing the docs and not understanding how it worked.  The callback gets executed too frequently.  Since it's used in the form helper, I decided it needed to change ;)
    - Sad face.  Docs required getting rid of `using` for eigen vector types :/

3. I'm going to split off onto another branch now for the remaining docs updates.  I made a serializer example, but that and the changes made for the docs will be in their own self-contained PR because they aren't exactly trivial changes.

[...]

PR #261, opened 2017-09-17
prevent serializer LTO symbols being stripped, example5

### LINKED ISSUE #210: Serialization of Themes
I see that Widgets have code for serialization using the Serializer (core.h/ serializer.cpp).
For me, I'd like create Themes for buttons, text box styles and save them.

I could go two ways:  1) save the themes in JSON, read them and create Themes on the fly, or 2) use the serializer to save/load themes, which I think stores in binary format.

Any comments or suggestions on the best way to approach this?

svenevs:
This can be merged uncorrelated to #260.  ~~If merged before, should apply cleanly on master.  If merging after, let me rebase (see notes about example5 and cmake).~~  On second thought, ideally get 260 in first and then I'll rebase.  I wanted these specific changes in a different PR because they're more than just simple fixes.

- LTO will strip out the serializer symbols because it is not explicitly used in the framework
    - dirty solution: force it back in the header
    - should fix #210
- example5 tests serialization
    - some widgets do not appear to serialize completely
        - e.g. what is selected for images or combobox
- no docs on example5 yet, would be done after merge of other docs PR (will create conflict because of CMake and formatting of examples.rst)
- changes in glutil.h
    - docs and anonymous enums create >50 additional warnings
    - C++11 to the rescue: static constexpr auto
    - minimal test with mSerialShader in example5
    - class `half_float::half` is a "problem"
        - no full definition given; can change to

              namespace half_float { class half {}; }

          ???  This will make it an opaque type that does nothing, but still allow template instantiation using it


svenevs:
(relaunch Travis, passed on fork, this failed on installing python deps for gcc)

src/example5.cpp
```diff
@@ -0,0 +1,247 @@
+/*
+    src/example5.cpp -- example of how to use the same serialzer to save the
```
wjakob: typo: serializer

src/example5.cpp
```diff
@@ -0,0 +1,247 @@
...
+#include <nanogui/serializer/core.h>
+#include <nanogui/serializer/opengl.h> // unused in this example
+#include <nanogui/serializer/sparse.h> // unused in this example
+
+#include <map>
+
+/*
+ * NOTE: not everything appears to get saved / restored.  The image selected,
+ *       what the combobox has selected, etc.  Feel free to investigate and
+ *       submit a PR if those clases are not serializing fully.  The original
```
wjakob: Last sentence can probably go ;)

wjakob:
I'm not sure I follow why the ``Serializer`` class definitions had to be moved to the header file. Isn't it enough to mark it as an export to ensure that the symbols end up in the .so file?

svenevs:
I believe it's because of LTO. I also believe it's either apple clang or llvm specific. On osx if you comment out all of the cmake stuff pertaining to LTO the serializer symbols will exist.

I've experienced a similar bug in my own project that enumerates template class instantiation in a source file. If I use the host compiler and NVCC, the symbols exist. If I use LLVMm too compile the cuda code, it'll stop them out because it sees that the library doesn't internally use the symbol anywhere and so it deleted it.

svenevs:
Docs updated now that the other one is merged in.  There is only one file changed here that I missed in the other PR, which is `VScrollPanel`.  All remaining docs are on a different PR.

See `examples.rst` changes.  In previous one I had been writing links that always use `nanogui::ClassName`.  That's going to be undone because it conflicts stylistically with the Breathe auto-links.  For example, if you just write `/// Overrides Widget` breathe will auto-link to `Widget` without a `nanogui::` prefix.  So the `examples.rst` "rst" links have been updated to accomodate that, but a different PR in the future will have those adjusted.

wjakob:
I suspect that all of the LTO-related issues are simply due to the fact that the class declaration of the serializer reads
```
class Serializer {
```

and not

```
class NANOGUI_EXPORT Serializer {
```

Is that possible?

svenevs:
Holy damn.  Fixed in #313.

PR #262, opened 2017-09-19
Fixed potential crashes caused by dangling pointers.

LEChaney:
In the current version, mDragWidget is a raw pointer, which can cause crashes if the widget it points is removed. mDragWidget is set on mouse click, and if the same click also removes the widget, a crash occurs when the mouse is released due to a dangling pointer.

### DIFF
--- a/include/nanogui/screen.h
+++ b/include/nanogui/screen.h
@@ -195,7 +195,7 @@ protected:
     int mMouseState, mModifiers;
     Vector2i mMousePos;
     bool mDragActive;
-    Widget *mDragWidget = nullptr;
+    ref<Widget> mDragWidget;
     double mLastInteraction;
     bool mProcessEvents;
     Color mBackground;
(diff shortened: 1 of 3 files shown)


wjakob:
I feel that there are a few redundant changes here. Why doesn't using ``ref<>`` alone fix the issue?

wjakob:
The ``if (mDragActive && mDragWidget->getRefCount() == 1) {`` logic in particular looks suspect/undesirable.

FurryFur:
Yes, ref alone does fix the issue. Ideally, these kinds of dangling pointers should probably be resolved by implementing some kind of weak reference, so I was trying to kind of lazily fudge the same functionality here since ref doesn't support weak references. The only issue with using ref alone is that the object is kept alive by the mDragWidget, even though it becomes invisible to the user. I thought this might perhaps cause unexpected behavior in some rare circumstances. In my tests, I found that the Widgets weren't destroyed until I released the mouse button without the extra checks. It seems like mDragWidget triggers an event on itself before it is finally destroyed, so I wanted to avoid this.

FurryFur:
I can certainly remove the extra checks if you would prefer. I also copied the mFocus dangling pointer fix from #198, but without any of the extra changes.

PR #263, opened 2017-09-23
fix tabheader right icon position

svenevs:
I'm having a really hard time debugging this, since the original works for me on OSX Sierra.  It's clearly got something to do with the positioning code.

Printout when I run it:

```
------------------------------
left pos:  4, 10
right pos: 224, 10
prev rpos: 221, 10
mPos:      0, 0
mSize:     240, 20
r width:   15
tab width: 20
------------------------------
```

`right pos` is the proposed new calculation that is currently getting applied.  `prev rpos` is what it used to be.  Can you run this and include the output / indicate whether the new positioning scheme is working for you?  I can't really reason about why it's breaking...the only thing I can think of is possible bad `float` / `int` conversion with the multiplication?

### DIFF
--- a/src/tabheader.cpp
+++ b/src/tabheader.cpp
@@ -443,8 +443,9 @@ void TabHeader::drawControls(NVGcontext* ctx) {
     nvgTextAlign(ctx, NVG_ALIGN_LEFT | NVG_ALIGN_MIDDLE);
     float yScaleRight = 0.5f;
     float xScaleRight = 1.0f - xScaleLeft - rightWidth / theme()->mTabControlWidth;
-    auto leftControlsPos = mPos.cast<float>() + Vector2f(mSize.cast<float>().x() - theme()->mTabControlWidth, 0);
-    Vector2f rightIconPos = leftControlsPos + Vector2f(xScaleRight*theme()->mTabControlWidth, yScaleRight*mSize.cast<float>().y());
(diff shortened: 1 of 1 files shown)


wjakob:
Huh, strange -- it's working perfectly on my machine at home.
```
------------------------------
left pos:  4, 10
right pos: 224, 10
prev rpos: 1, 10
mPos:      0, 0
mSize:     240, 20
r width:   15
tab width: 20
------------------------------
```
I will try again on my office PC next week, that's where I noticed the issue.

svenevs:
> `prev rpos: 1, 10`

Actually that's suspect.  I think it's a casting problem.  After testing in-office, if it positions correctly then the "new" calculation should be the fix (in which case I'll fix this up to just use that).

wjakob:
This can be closed, right?

svenevs:
I guess that depends on whether or not the icon position on `master` is still broken for you...if it is and this branch works, the "new" (refactored but basically the same) computation should be incorporated.  As a refresher, the right icon position for the tab widget in example 1 is the problem.

wjakob:
Just to confirm (after testing on my office machine): it's still broken on master, and it works with this PR.

svenevs:
Updated, it turns out after upgrading to wayland I had the same problem only on linux, this fixes it though.

wjakob:
Merged, thank you!

Issue #202, opened 2017-01-23
Support for high DPI screens


rogerdahl:
Thank you for NanoGUI, it's awesome!

Are there any plans to add support for scaling on Linux? I wrote an app using NanoGUI and it uses the size passed to Screen directly as pixel size. After upgrading my monitor to 4K and increasing the OS scale factor to match, most apps and GUI elements scaled appropriately, but my NanoGUI app is now tiny...

svenevs:
Are you using [`Screen::pixelRatio()`](http://nanogui.readthedocs.io/en/latest/generated_api/exhale_class_nanogui__Screen.html#_CPPv2NK7nanogui6Screen10pixelRatioEv) (or its equivalent `mPixelRatio` member variable)?  You should be able to use it to multiply with [`Widget::mSize`](http://nanogui.readthedocs.io/en/latest/generated_api/exhale_class_nanogui__Widget.html#_CPPv2N7nanogui6Widget5mSizeE), but you may have to deal with `float` vs `int` stuff at some point too.

rogerdahl:
I checked now and found that `Screen::pixelRatio()` return `1`.

svenevs:
That's unfortunate :/  Out of curiosity, if you manually resize your screen class to scale everything by 2 does it appear as you desire?  (by 2 to go from 1920x1080 to 3840x2160).

If it does, then I'm not sure what to say.  At the top of [`screen.cpp`](https://github.com/wjakob/nanogui/blob/master/src/screen.cpp#L47) is where this is getting setup, I'm inclined to assume you may not have `gtk` or something.  One thing you could try is getting rid of the `__linux__` part so that it looks like

```
/* Calculate pixel ratio for hi-dpi devices. */
static float get_pixel_ratio(GLFWwindow *window) {
#if defined(_WIN32)
    HWND hWnd = glfwGetWin32Window(window);
    HMONITOR monitor = MonitorFromWindow(hWnd, MONITOR_DEFAULTTONEAREST);
    /* The following function only exists on Windows 8.1+, but we don't want to make that a dependency */
    static HRESULT (WINAPI *GetDpiForMonitor_)(HMONITOR, UINT, UINT*, UINT*) = nullptr;
    static bool GetDpiForMonitor_tried = false;

    if (!GetDpiForMonitor_tried) {
        auto shcore = LoadLibrary(TEXT("shcore"));
        if (shcore)
[...]

PR #265, opened 2017-09-27
Fix for scaling on high-dpi screens when using KDE

dvicini:
As discussed in 
https://github.com/wjakob/nanogui/issues/202
This fixes the UI scaling for systems which use KDE5.

### DIFF
--- a/src/screen.cpp
+++ b/src/screen.cpp
@@ -485,7 +485,7 @@ bool Screen::cursorPosCallbackEvent(double x, double y) {
     Vector2i p((int) x, (int) y);
 
 #if defined(_WIN32) || defined(__linux__)
-    p /= mPixelRatio;
+    p = (p.cast<float>() / mPixelRatio).cast<int>();
 #endif
 
     bool ret = false;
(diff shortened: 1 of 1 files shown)


dvicini:
This does not support fractional pixel scaling yet. Let me first fix that before this is merged...

src/screen.cpp
```diff
@@ -472,7 +485,7 @@ bool Screen::cursorPosCallbackEvent(double x, double y) {
     Vector2i p((int) x, (int) y);
 
 #if defined(_WIN32) || defined(__linux__)
-    p /= mPixelRatio;
+    p = (p.cast<float>() / mPixelRatio).cast<int>();
```
svenevs: does it make sense to "round"?  `p = ((p.cast<float>() / mPixelRatio) + 0.5f).cast<int>();`

I'm not sure I understand the impact of fractional computations since it's getting turned into an `int` anyway...

Tom94:
Fractional pixel scaling on Windows will still not work with your changes as the obtained pixel ratio is still being rounded to an integer in line 65. Simply removing the `round` call should be sufficient to fix it.

src/screen.cpp
dvicini: So the rounding in the end (`.cast<int>()`) is necessary, as Eigen requires to explicitly convert between the different types and `p` is an int-vector. Does that answer your question?

dvicini:
Thanks Thomas, I removed the rounding in Line 65 to also fix the scaling on Windows.

wjakob:
Thanks, this looks great!

PR #272, opened 2017-10-17
Add setSizeLimits

DomainsHax:


src/screen.cpp
```diff
@@ -369,6 +369,10 @@ void Screen::setSize(const Vector2i &size) {
 #endif
 }
 
+void Screen::setSizeLimits(const Vector2i& minsize, const Vector2i& maxsize) {
+  glfwSetWindowSizeLimits(mGLFWWindow, minsize[0], minsize[1], maxsize[0], maxsize[1]);
```
wjakob: Please use 4 spaces for indentation.

wjakob:
This looks reasonable, but the function you added is missing Python bindings. Can you please add them?

PR #273, opened 2017-10-18
example1 can not start because icons directory is not found

rblath:


svenevs:
Hi @rblath,

What OS are you on and what are the sequence of steps you took to build this?  On Unix systems, the expected workflow is to build out of source, one directory below.

```
nanogui/
    include/
    src/
    ...
    build/
        * you build here *
```

What were the steps you took that led you to need `..`?

wjakob:
It's just a matter of starting nanogui from the right directory as @svenevs says, so I don't think the proposed patch makes sense -- I'll go ahead and close this ticket.

rblath:
Sorry that I did not get around to answer this sooner. The build is made on Windows 7 with VS 2017. 
The steps are:

git clone --recursive https://github.com/wjakob/nanogui.git
cd nanogui
mkdir build
cd build
cmake ..
msbuild ALL_BUILD.vcxproj /p:Configuration=Release

Now here is where I usually do "cd Release" and then start "example1.exe" which then fails because it does not find the icons directory. That's why I made the pull request.
But I acknowledge that one can just as easily do:

Release\example1.exe

And the example starts without problems. So if you don't want to take the PR then that's fine by me.

Thanks for looking into this.

Rene

PR #281, opened 2017-11-09
allow const acquisition of Arcball's Quaternionf

svenevs:
Needed for a scenario such as

```cpp
class ArcballCamera {
    //                               vvvv
    Eigen::Quaternionf getRotation() const {
        return mArcball.state();
    }
protected:
    nanogui::Arcball mArcball;
}
```

Maybe the desired signature would be `const Quaternionf &getState() const { return mQuat; }`?  The `const` part is the main problem, ref vs stacked quat aren't really a concern.

### DIFF
--- a/include/nanogui/glutil.h
+++ b/include/nanogui/glutil.h
@@ -465,6 +465,8 @@ struct Arcball {
 
     Quaternionf &state() { return mQuat; }
 
+    const Quaternionf &state() const { return mQuat; }
+
     void setState(const Quaternionf &state) {
         mActive = false;
         mLastPos = Vector2i::Zero();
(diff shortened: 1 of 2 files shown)


include/nanogui/glutil.h
```diff
@@ -465,6 +465,8 @@ struct Arcball {
 
     Quaternionf &state() { return mQuat; }
 
+    Quaternionf getState() const { return mQuat; }
```
wjakob: Please use the same naming conventions:
``const Quaternionf &state() const { return mQuat; }``

If you chose this different name to avoid an error in the python bindings, you'll need to use ``py::overload_cast<>(&Quaternionf::state)``

wjakob:
Ah right, there were some issues with MSVC2015. The legacy variant is of course also good.

PR #282, opened 2017-11-12
Remaining fonts bypass

svenevs:
Allows for font bypassing of all widgets that currently have hard-coded font names: `Button`, `CheckBox`, `Graph`, `ImageView`, `Screen`, `TextBox`, and `Window` (first commit accidentally omitted `TextBox`, found that when doing the bindings).

I finally decided to try using a new font face and discovered it was only possible with selected widgets.  In my particular use case, I needed to distinguish a capital `I` from a lower case `l` on a `Button`.  Getting the font into `nanogui/resources` is easy enough (advanced compilation docs), but using it is currently not possible.

**Before Merge**:

- [ ] (me) add a better description of how to use a custom theme in a parent project.  The best approach for a universal theme change is by setting the theme of the `Screen` class before adding children (rather than setting the theme for every child).
- [x] (@wjakob) decision on monospace font being included in the upstream
- [x] (me) figure out how to properly overload default fonts
- [ ] move the font size stuff to `FontWidget`
- [ ] this needs both a re-base and some cherry picking, so for simplicity i'm going to close this PR and open a new one when it's ready since this got out of hand.

For the second one, I think it makes sense to vendor `RobotoMono-{Regular,Bold}` with NanoGUI.  It's the same license (Apache2).  It's nice because it has that beautiful slash through the `0`, attached are some screenshots of its would-be usage.  Right now this is **not** in this PR.  It could be, and the diff is quite small.  Two more fonts loaded in `Theme`, and I made `mFont` for `FloatBox` and `IntBox` default to `"mono"` (rather than parent class `TextBox` using `"sans"`).  **Note** the spacing around the `.` in the `FloatBox` / `IntBox`.  That may not be something you like.
    - Side question: if including monospace, should `ImageView` default to `"mono"`?

I think a regular and monospace font are sufficient, and I could rewrite the docs on how to do fonts to show an example of an italic font face if a user really needs it.  Otherwise, if you'd rather not have the monospace fonts, I'll tailor the docs to show how to use a monospace font :wink:

mono_icons_demo

mono_example_1

mono_example_2


wjakob:
This looks great, and I have no objections to including the monospace font. I don't think it's a good default for Float/IntBox though, would you mind reverting to the previous "sans"? We could still have a widget in "example1" where the mono font is explicitly selected.

svenevs:
Cool!  I had wanted to fix that grammar problem a long time ago but kind of forgot.  Added a monospace `IntBox` just below the existing (`"sans"`, no widgets default to `"mono"`).  The window below it for the color picker got bumped down a little as a result.  I think it looks reasonable, but can change it.

random_mono
 
So in terms of the documentation, what are your thoughts on me encouraging people to `#include <nanogui_resources.cpp>` (the **.cpp** file) directly?  The terse version for a theme that wants a custom font:

1. CMake side: copy as already instructed, but also add

    ```cmake
    # Using a custom nanogui theme, we need the nanogui_resources.{h,cpp} available
    list(APPEND DSCAN_EXTRA_INC_DIRS ${CMAKE_CURRENT_BINARY_DIR}/nanogui)
    ```

2. Create a header `my_theme.h` that includes `<nanogui/theme.h>`, create derived class, only provide constructor stub to be implemented in `.cpp` file.

3. (necessary) Create `my_theme.cpp` which includes `my_theme.h`, and **also** `nanogui_resources.cpp`.

LTO will (correctly) strip out the font resources the user copied in with (1) from the `nanogui` library, so you will fail at link time.  By making a terminal `my_theme.cpp` and including `nanogui_resources.cpp`, you get them back again.  It's kind of dubious, but I don't know how else you would do it.

wjakob:
I'm not sure I followed your 3-step explanation above. Is the goal here to include even more fonts on the user side? Why would this not work just using the default mechanisms of registering fonts, etc.?

svenevs:
> Is the goal here to include even more fonts on the user side?

Correct, e.g. an italic or sans-serif font or something else they wanted to copy in.  CMake can be used to copy the font and thereby NanoGUI will create the `font_name_ttf` and `font_name_ttf_size` in both `nanogui_resources.h` and `nanogui_resources.cpp` in the binary directory.

> Why would this not work just using the default mechanisms of registering fonts, etc.?

The issue is that currently if you add a custom font by copying it to `nanogui/resources`, everything will get generated, but at link time (if you only include `nanogui_resources.h`) you will have undefined symbols.  I believe it's because NanoGUI does not use the definitions, so they just get stripped out.

```
libfont-demo.so: undefined reference to `spirax_regular_ttf_size'
libfont-demo.so: undefined reference to `spirax_regular_ttf'
```

I made a really dirty demo repo, toggle [commenting `.h` vs `.cpp` include to reproduce](https://github.com/svenevs/nanogui-custom-font-demo/blob/57e34943fe37ee8d1427211c35868c0cf2951937/src/my_theme.cpp#L5-L6).

I'm kind of surprised this tactic doesn't lead to collisions during the link of `libfont-demo` and `libnanogui` itself though.  By including the `.cpp` that NanoGUI used for the Roboto fonts...shouldn't it show up twice?  I didn't do any fancy LTO or anything in that project.

wjakob:
Another alternative would be to have a CMake parameter, e.g. ``NANOGUI_EXTRA_RESOURCES``. Any files added to this list (e.g. by a parent cmake project) could be passed through bin2c as well, with some minor name mangling to produce a suitable symbol name.

The loss of symbols is probably just a visibility-related issue. I suspect that they just need a ``NANOGUI_EXPORT`` so that the linker knows that it should make an entry in the symbol table.

svenevs:
Ok yeah the `NANOGUI_EXTRA_RESOURCES` approach is pretty nice, it allows the NanoGUI build system to automatically include the binary directory where the generated resources get put.  Where symbols go, visibility is not enough.  It's basically the same problem as the `Serializer` symbols -- NanoGUI is not using the custom font symbols anywhere, so they get stripped out.

What are your thoughts on making some kind of static map of sorts that is hidden in `nanogui::detail` and clearly documented as never use?  Or the map could actually get used in `theme.cpp`?  It seems like overkill, but we need to engineer a way to force LTO to keep the symbols.  My understanding of the general state of affairs is that `__attribute__((used))` is at best a suggestion, despite it being the officially documented way of keeping it.

svenevs:
### What is working

Parent projects simply need to

```cmake
list(APPEND NANOGUI_EXTRA_RESOURCES "${CMAKE_CURRENT_SOURCE_DIR}/resources/spirax/Spirax-Regular.ttf")
list(APPEND NANOGUI_EXTRA_RESOURCES "${CMAKE_CURRENT_SOURCE_DIR}/resources/membra/membra.ttf")
# etc
```

See changes in `bin2c.cmake`, they're significant.  The `std::map` with associated lookup function `nanogui::createFontMem` is what guarantees LTO will not remove the symbols.

### Fatal flaw

- Default fonts `mFontNormal` and `mFontBold` cannot be overridden.
    - It seems this is currently impossible, as [fonts cannot be freed](https://github.com/memononen/nanovg/issues/337).
    - Implication: in a derived class

        ```cpp
        mFontNormal = nanogui::createFontMem(ctx, "sans", "SpectralSC-Regular.ttf");
        mFontBold = nanogui::createFontMem(ctx, "sans-bold", "SpectralSC-Regular.ttf");
        ```

        the original Roboto fonts will still be used.

- Option 1: add some boolean parameters to the theme constructor
    - `Theme(NVGcontext *ctx, bool loadSans = true, bool loadMono = true, bool loadIcons = true)`
        - Kind of error prone, I also couldn't actually get it to work.  Probably could, but it felt really awkward.
- Option 2: since we've already moved the default widget icons (where applicable) into the `Theme` class, maybe there would be a clean way of reworking things to use theme-based font defaults.
    - Envisioned scenario in `Theme` class:

        ```cpp
        std::string mNormalFontName = "sans";
        std::string mBoldFontName = "sans-bold";
        ```

    - Primary problem is that things right now are setup with default arguments.  Example `Label`:

        ```cpp
        Label(Widget *parent, const std::string &caption,
              const std::string &font = "sans", int fontSize = -1);
        ```

        This isn't valid, but it would need to be something like

        ```cpp
        Label(Widget *parent, const std::string &caption,
              const std::string &font, int fontSize = -1);
        Label(Widget *parent, const std::string &caption)
            : Label(parent, caption, mTheme->mNormalFontName) { }
        //                           ^^^^^^^^^^^^^^^^^^^^^^^
        ```

Anybody have thoughts on how to re-design this?

svenevs:
Ok if anybody is following along on this, I think I've figured out the best way to do this.

### At Build Time

Specifying the fonts you want embedded [is easy](https://github.com/svenevs/nanogui-custom-font-demo/blob/aa9ad18cdb937b207a1b64b4e5c376345aaeabbf/CMakeLists.txt#L8-L11):

```cmake
list(APPEND NANOGUI_EXTRA_RESOURCES "${CMAKE_CURRENT_SOURCE_DIR}/resources/spirax/Spirax-Regular.ttf")
list(APPEND NANOGUI_EXTRA_RESOURCES "${CMAKE_CURRENT_SOURCE_DIR}/resources/membra/membra.ttf")
list(APPEND NANOGUI_EXTRA_RESOURCES "${CMAKE_CURRENT_SOURCE_DIR}/resources/spectral/SpectralSC-Regular.ttf")
list(APPEND NANOGUI_EXTRA_RESOURCES "${CMAKE_CURRENT_SOURCE_DIR}/resources/spectral/SpectralSC-Bold.ttf")
```

### Internal `std::map` of the fonts guarantees LTO won't strip it out

The generated `nanogui_resources.cpp` will do something like this

```cpp
std::map<const std::string, const std::pair<uint8_t *, uint32_t>> __nanogui_font_map = {
    {"Roboto-Bold.ttf", {roboto_bold_ttf, roboto_bold_ttf_size}},
    {"Roboto-Regular.ttf", {roboto_regular_ttf, roboto_regular_ttf_size}},
    {"RobotoMono-Bold.ttf", {robotomono_bold_ttf, robotomono_bold_ttf_size}},
    {"RobotoMono-Regular.ttf", {robotomono_regular_ttf, robotomono_regular_ttf_size}},
    {"entypo.ttf", {entypo_ttf, entypo_ttf_size}},
    {"Spirax-Regular.ttf", {spirax_regular_ttf, spirax_regular_ttf_size}},
    {"membra.ttf", {membra_ttf, membra_ttf_size}},
    {"SpectralSC-Regular.ttf", {spectralsc_regular_ttf, spectralsc_regular_ttf_size}},
    {"SpectralSC-Bold.ttf", {spectralsc_bold_ttf, spectralsc_bold_ttf_size}}
};

int createFontMem(NVGcontext *ctx, const std::string &fontName, const std::string &fontSource) {
    if (__nanogui_font_map.count(fontSource) > 0) {
        const auto &p = __nanogui_font_map[fontSource];
        return nvgCreateFontMem(ctx, fontName.c_str(), p.first, p.second, 0);
    }
    return -1;// signals error
}
```

### Final Tension

The final problem was the tension between how fonts are created, and how fonts are used.  They are created in `nanogui::Theme` or a derived class, but then different widgets such as `nanogui::Window` or `nanogui::Label` maintain their own `mFont` definition.  Adding [these methods to `nanogui::Theme`](https://github.com/svenevs/nanogui/blob/8da408c4db327878b01da5a3d1251f2d0e6af1da/include/nanogui/theme.h#L45-L51) allows for [`FontWidget::setDefaultFont`](https://github.com/svenevs/nanogui/blob/8da408c4db327878b01da5a3d1251f2d0e6af1da/src/fontwidget.cpp#L40).

Now, instead of `const std::string &font = "sans-bold"` for the `nanogui::Window` constructor, it is the empty string.  The empty string specifies that defaults should be chosen, and later that if `setTheme()` is called, the new theme's defaults should be used.  On the other hand, if a user does anything to set the font explicitly, theme setting will not change it.

Anybody have thoughts on this approach?  I'll be translating `nanogui::Button`, `nanogui::Label` etc to those soon.  If done right, zero external api changes are possible, but internally anything dealing with drawing text will inherit from `FontWidget`.

svenevs:
Ok so I've got it all mostly finished, another problem arose though.  My belief is the reason `nanogui::Theme` is not bound in python is because that would require creating bindings for `NVGcontext`.  Along the same vein, in the new `nanogui_resources.h`:

```cpp
/* Pass-through method to call ``nvgCreateFontMem``.
 *
 * Parameters:
 *     - ctx: the NanoVG context (included in nanogui::Theme constructor).
 *     - fontName: the name to register (e.g., "sans" or "sans-bold").
 *     - fontSource: the *basename* of the font source (e.g., "Roboto-Regular.ttf").
 */
extern NANOGUI_EXPORT int createFontMem(NVGcontext *ctx, const std::string &fontName, const std::string &fontSource);
```

cannot be bound.  I was looking at the widget overloads, and there are bindings, but I don't think they are valid:

```
>>> w = nanogui.Window(s, "window") # s is a nanogui.Screen
>>> w.theme()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: Unable to convert function return value to a Python type! The signature was
	(self: nanogui.Widget) -> nanogui::Theme
```

Is there a way around this without essentially binding all of `NanoVG`?  I have little interest in doing that, but without it custom themes / fonts are not available in Python (only C++).

PR #285, opened 2017-11-20
Glutil permissions and some docs fixes

### LINKED ISSUE #284: Arcball and GLShader -- too protected?
There are two things I constantly modify by inheriting from these classes and providing public access.  Both are simple enough that it really makes no difference to me whether or not they go in the upstream.  I just use them often enough that they may be good candidates.

### `Arcball`

This one may be a good candidate to just add to the `Arcball` class, without needing to expose anything:

```cpp
struct InterruptableArcball : public nanogui::Arcball {
    InterruptableArcball(float speedFactor = 2.0f) : Arcball(speedFactor) { }
    InterruptableArcball(const Quaternionf &quat) : Arcball(quat) { }

    void interrupt() {
        Arcball::button(mLastPos, false);
    }

public:
    EIGEN_MAKE_ALIGNED_OPERATOR_NEW
};
```

It's useful to be able to "close" the state without an actual mouse release.  I use the `Arcball` to model a simple camera model for interacting with the scene, but if an event occurs that issues say a `MessageDialog`, or really any other widgets become in focus, I can simply call `interrupt` to prevent some glitchy re-interaction (since `mLastPos` and the new click location may be very different).

### `GLShader`

Don't make the `GLShader::Buffer` or `GLShader::mBuffer` `protected`.  This one is a little more specialized.  Docs could be added stating plainly "you break it you buy it" -- users who choose to directly read / modify these values should be cautious and well...really need to know what they're doing.

The standard use case for me is CUDA / OpenGL interop.  If I want to write to the positions of a shader or something, I need the specific `GLuint` associated with this attribute to map a pointer with CUDA.  The class I wrote a long time ago that really doesn't need a template, but has it anyway:

```cpp
class ExposedGLShader : public nanogui::GLShader {
protected:
    enum class BufferAttribs : uint8_t {
        id = 0,
        glType,
        dim,
        compSize,
        size
        // version already has a getter
    };

    template <BufferAttribs type>
    GLuint *buffer_internal(nanogui::GLShader::Buffer &buff) {
        if(type == BufferAttribs::id)            return &buff.id;
        else if(type == BufferAttribs::glType)   return &buff.glType;
        else if(type == BufferAttribs::dim)      return &buff.dim;
        else if(type == BufferAttribs::compSize) return &buff.compSize;
        else if(type == BufferAttribs::size)     return &buff.size;

        return nullptr;
    }

    template <BufferAttribs type>
    GLuint* buffer_attrib(const std::string &name, bool warn = true) {
        for(auto &pair : mBufferObjects) {
            if(pair.first == name)
                return buffer_internal<type>(pair.second);
        }

        if(warn)
            std::cerr << mName << ": warning: did not find attrib " << name << std::endl;
        return nullptr;
    }

public:
    GLuint *bufferId(const std::string &name) {
        return buffer_attrib<BufferAttribs::id>(name);
    }

    GLuint *bufferGLType(const std::string &name) {
        return buffer_attrib<BufferAttribs::glType>(name);
    }

    GLuint *bufferDim(const std::string &name) {
        return buffer_attrib<BufferAttribs::dim>(name);
    }

    GLuint *bufferCompSize(const std::string &name) {
        return buffer_attrib<BufferAttribs::compSize>(name);
    }

    GLuint *bufferSize(const std::string &name) {
        return buffer_attrib<BufferAttribs::size>(name);
    }
};
```

Upstream fixes would either be exposing the `GLShader::Buffer` class, or possibly just adding the `ExposedGLShader` class.  Or neither.

Let me know if you think either of these are desirable.

svenevs:
Fixes #284.

### `Arcball`

- Docs completed, including brief example of how to coordinate it with a `nanogui::Screen`.
- Added `Arcball::interrupt`, which just calls `button(Vector2i::Zero(), false)` to close out state in absense of mouse click.
- Added the ability to gain the active rotation as a quaternion (`Quaterionf Arcball::activeState() const`).  Currently it can only be obtained with `Arcball::matrix`, which is nice for traditional model-view-projection creation.  Having access to the updated quaternion is useful for more efficient / convenient for direct `Vector3f` transformations.
- Associated python bindings / `py_doc.h` added.

### `GLShader`

- Docs completed.
- `GLShader::Buffer` is now a `public` nested type.
- Acquisition of e.g., `"position"` can be done with the new `const GLShader::Buffer & attribBuffer(const std::string &name) const`.
    - Throws `std::runtime_error` if not found.
    - Example usage for CUDA would be

        ```cpp
        // map the positions to write from CUDA
        const auto &position_buff = mPointShader.attribBuffer("position");
        GLuint gl_positions = position_buff.id;
        cudaGraphicsGLRegisterBuffer(
            &mPositionsCudaResource, gl_positions, cudaGraphicsRegisterFlagsWriteDiscard
        );
        ```
- `nanogui.Buffer` class and `nanogui.GLShader.attribBuffer` are bound, but honestly I don't know how useful they will be on the python side.

### Docs Fixes

- I had some really crazy issues with the `pybind11/tools/clang`, which resulted in large amounts of docs being deleted (exceptions or occasional segfault (?!)).
    - Almost certainly a local problem, but enabled users to e.g., `pip3 install clang-5` if they experience the same problem.
    - The majority of the changes in `docs/mkdoc_rst.py` are accommodating this.
        - Try and import first, then fallback on `pybind11` vendored copy.
        - Changed `d(s)` to enable using `clang-5`, theirs is returning strings which can't be decoded.
- Make sure `DOXYGEN_DOCUMENTATION_BUILD` is defined (via `CMakeLists.txt`) when executing `make mkdoc`.

### DIFF
--- a/docs/mkdoc_rst.py
+++ b/docs/mkdoc_rst.py
@@ -157,6 +157,8 @@ def process_comment(comment):
             s = s[2:].lstrip('*')
         elif s.endswith('*/'):
             s = s[:-2].rstrip('*')
+        elif s.startswith('///<'):
+            s = s[4:]
         elif s.startswith('///'):
             s = s[3:]
         if s.startswith('*'):
@@ -207,8 +209,8 @@ def process_comment(comment):
         'sa': 'See also',
         'see': 'See also',
         'extends': 'Extends',
-        'throw': 'Throws',
-        'throws': 'Throws'
+        'throws': 'Throws',
+        'throw': 'Throws'
     }.items():
         s = re.sub(r'\\%s\s*' % in_, r'\n\n$%s:\n\n' % out_, s)
 
--- a/docs/mkdoc_rst.py
+++ b/docs/mkdoc_rst.py
@@ -257,12 +257,10 @@ def process_comment(comment):
                 if len(line) < 4:
                     result += line.strip()
                 else:
-                    # this is a .. code-block:: indentation (three spaces)
+                    # this is a .. code-block:: indentation (three spaces),
+                    # strip leading three spaces, preserve remaining indentation
                     if line.startswith('   '):
-                        if line[3] != ' ': # outer-most indentation level
-                            result += line[3:].strip() + '\n'
-                        else:
-                            result += line[3:].rstrip() + '\n'
+                        result += line[3:].rstrip() + '\n'
                     else:
                         result += line.strip() + '\n'
         else:
--- a/docs/mkdoc_rst.py
+++ b/docs/mkdoc_rst.py
@@ -14,6 +14,7 @@
 #        s = re.sub(r'\\endrst', r'```\n\n', s)
 #        s = re.sub(r'.. note::', r'Note:', s)
 #        s = re.sub(r'.. warning::', r'Warning:', s)
+#        s = re.sub(r'.. danger::', r'Danger:', s)
 #        s = re.sub(r'.. code-block::\s*\w*', r'', s)
 #
 #  - process_comment: near the end, have to treat .. code-block:: specially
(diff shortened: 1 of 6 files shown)


docs/mkdoc_rst.py
```diff
@@ -230,8 +258,11 @@ def process_comment(comment):
                     result += line.strip()
                 else:
                     # this is a .. code-block:: indentation (three spaces)
-                    if line.startswith('   ') and line[3] != ' ':
-                        result += line[3:].strip() + '\n'
+                    if line.startswith('   '):
+                        if line[3] != ' ': # outer-most indentation level
```
svenevs: I guess maybe this could be simpler, just change to

```py
if line.startswith('   '):
    result += line[3:].rstrip() + '\n'
```

?

include/nanogui/glutil.h
```diff
@@ -520,18 +736,57 @@ struct Arcball {
...
+     *        \ref Arcball::button with \ref mLastPos and ``false``.
+     *
+     * Use this method to "close" the state of the Arcball when a mouse release
+     * event is not available.  You would use this method if you need to stop
+     * the Arcball from updating its internal rotation, but the event stopping
+     * the rotation does **not** come from a mouse release.  For example, you
+     * have a callback that created a \ref nanogui::MessageDialog which will now
+     * be in focus.
+     */
+    void interrupt() { button(mLastPos, false); }
```
svenevs: Insofar as I understand, it truly doesn't matter if we use `button(mLastPos, false)` since it is a "release".  However, would it be more correct to do `button(Vector2i::Zero(), false)` instead?

svenevs:
Ok this should fix everything, noting that there is one item that was never discussed on the other issue: `Quaterionf Arcball::activeState()`.  See revised comment above.

svenevs:
@wjakob have you had time to review this / things need to change, or not had time to review yet?  This just got rebased onto master.

wjakob:
This looks fantastic, thank you!

PR #288, opened 2017-11-29
focusEvent when Screen (application) out of focus

svenevs:
The `Widget::focusEvent` handler was already in place, this just makes it so that the `Screen` gets notified in the case that the application goes out of focus.  See [glfw docs on input focus](http://www.glfw.org/docs/latest/window_guide.html#window_focus).  I don't think there are any gotchas in terms of the other widgets and their focus, it should be agnostic to this.

In one of my applications I was stealing the mouse after a key press, but without an event notifying the application itself is out of focus, if I alt+Tab away I still have no mouse!

### DIFF
--- a/src/screen.cpp
+++ b/src/screen.cpp
@@ -255,16 +255,28 @@ Screen::Screen(const Vector2i &size, const std::string &caption, bool resizable,
        a window from a Retina-capable screen to a normal
        screen on Mac OS X */
     glfwSetFramebufferSizeCallback(mGLFWWindow,
-        [](GLFWwindow* w, int width, int height) {
+        [](GLFWwindow *w, int width, int height) {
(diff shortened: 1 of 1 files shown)


wjakob:
Embarrassed how old this PR is (I was swamped with the SIGGRAPH deadline) -- would you mind changing the ``static_cast<bool>`` to a comparison (i.e. ``!= 0``). Some compilers, particularly MSVC warn about casts from integers to booleans. Naturally the code that is emitted in the end is identical.

svenevs:
No worries.  The SIGGRAPH deadline is at such an unfortunate time of year :S

Removed the `static_cast`, this also got rebased onto the current master.

wjakob:
Thanks!

PR #294, opened 2017-12-23
Maximize option was added

jkvargas:
Just adding an option to maximize the window.


svenevs:
Cool!  The [`Screen` python bindings](https://github.com/wjakob/nanogui/blob/3e0b2c8dae3a52d88d902d1ef65a86258be00c38/python/widget.cpp#L98-L101) need to be updated, you would add `py::arg("maximized") = false` after `"glMinor"` but before the `D(Screen, Screen)`.

Please also add documentation for this in the header file briefly explaining what the `maximized` parameter does :)

jkvargas:
Sure svenevz, during this week i'll do for sure

Issue #267, opened 2017-10-04
won't generate *.dylib on macOS with Xcode


dingzeyuli:
I tried to compile nanogui library on my mac. I use cmake to generate an Xcode project. The `libglfw_objects` and `nanogui-obj` can be generated. But the final `nanogui` does not show up.

After some digging, I think this might be the reason. 

> Some native build systems may not like targets that have only object files, so consider adding at least one real source file to any target that references `$<TARGET_OBJECTS:objlib>`.

Sources: [[1](https://gitlab.kitware.com/cmake/cmake/issues/16524
)] [[2](https://cmake.org/cmake/help/v3.7/command/add_library.html?#object-libraries)]

My workaround is to add a blank.c file in the nanogui src and modify the CMakeLists.txt:
```
add_library(nanogui
${NANOGUI_LIBRARY_TYPE}
blank.c
$<TARGET_OBJECTS:nanogui-obj>
$<TARGET_OBJECTS:glfw_objects>
)
```
The blank.c file is empty, just there to activate the Xcode building system. With this modification, nanogui now compiles with Xcode.

svenevs:
Thanks for the detailed report @dingzeyuli, is this the same as #220?  That one kind of disappeared into the fold, but I believe it achieves the same fix.

dingzeyuli:
Yeah, thanks @svenevs. That pull request has not yet been accepted. So the current nanogui code still fails with Xcode+cmake. That one seems to pass the integration test. Is there any plan to accept it? Thank you.

wjakob:
That PR failed the CI tests, so I never seriously looked at it. I'll write some comments on #220.

liminchen:
Hi Liding, Is the problem you encountered the one below?
I tried your fix but it still doesn't work...
<img width="1136" alt="screenshot 2018-01-25 10 53 31" src="https://user-images.githubusercontent.com/6930359/35406606-3900e406-01be-11e8-8626-856dc58ed261.png">

svenevs:
Fixed in #304.

PR #304, opened 2018-01-27
fix XCode build

svenevs:
- Revives original solution proposed in #220
    - Thanks @AndySomogyi !!!
- Adds exclusion of GLFW library, XCode fails to link this.
    - @liminchen this should fix your [linking problems here](https://github.com/wjakob/nanogui/issues/267#issuecomment-360564228)

This worked for me on XCode 8, can other people test and make sure it works for them?  I was able to build everything and run the executables from the *terminal*, but I don't actually use XCode and couldn't figure out how to get a GLFW app to run from XCode (something about what the terminal is).

@dingzeyuli does this PR work for you?

To test:

```console
$ git clone --recursive https://github.com/svenevs/nanogui.git
$ cd nanogui
$ git checkout fix_xcode
$ mkdir build && cd build
$ cmake .. -G Xcode
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -341,10 +341,30 @@ add_library(nanogui-obj OBJECT
   src/serializer.cpp
 )
 
-add_library(nanogui ${NANOGUI_LIBRARY_TYPE}
-  $<TARGET_OBJECTS:nanogui-obj>
-  $<TARGET_OBJECTS:glfw_objects>
-)
+# XCode has a serious bug where the XCode project produces an invalid target
+# that will not get linked if it consists only of objects from object libraries,
+# it will not generate any products (executables, libraries). The only work
+# around is to add a dummy source file to the library definition. This is an
+# XCode, not a CMake bug. See: https://itk.org/Bug/view.php?id=14044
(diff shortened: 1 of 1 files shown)


dingzeyuli:
Yes. This PR fixes the problem I had with dummy empty files. The code works on my machine.

Thanks @svenevs.

wjakob:
Looks good to me, what a weird XCode bug -- thanks for tracking that down!

liminchen:
Thank you, svenevs! It also fix the problem on my machine.

svenevs:
Cool, thanks for testing everybody!

@wjakob the symptoms of the GLFW link problem ultimately stem from the same issue (`glfw_objects` library, needs the dummy empty file).  It may be slightly more ideal to incorporate that in the patched build system if GLFW gets updated.  On the other hand, there's actually no need to link GLFW at all -- for any build system.  So maybe it should just be skipped for everything?

wjakob:
I agree that GLFW should probably not be compiled separately if it's used via glfw_objects in a parent project. Would you be willing to submit a PR that makes this change?

svenevs:
Sure thing, coffee first though

PR #305, opened 2018-01-28
skip linking GLFW for all build systems

svenevs:
Instead of skipping just for Xcode, skip for all.  See [discussion here](https://github.com/wjakob/nanogui/pull/304#issuecomment-361069171).

This is particularly nice for parent projects that set `CMAKE_*_OUTPUT_DIRECTORY`, as GLFW will no longer appear.  Worked on OSX (Xcode and Makefile) and Linux (Makefile).

Assuming the Appeveyor tests work on windows, it should be good!


wjakob:
Merged!

PR #306, opened 2018-02-01
prevent children from being added to TabWidget

### LINKED ISSUE #303: floating point exception for TabWidget
I was creating a tab widget and wasn't being very careful, it looked something like this

```cpp
mTabWidget = new TabWidget(mControlsWindow, "Controls");
auto *layer = mTabWidget->createTab("First");
// ... some fancy layout stuff that rely on wrappers ...
//                         vvvvvvvvvv
auto *wrapper = new Widget(mTabWidget);// !!!!!
//                         ^^^^^^^^^^
wrapper->setLayout(new GroupLayout());
```

The `wrapper` needed to have `layer` not `mTabWidget` as the parent, but the side effect is that when the tab widget performs its layout a floating point exception occurs.

Can I just add docs to the tab widget saying "**never** add a child directly to the tab widget, **always** create a new tab and add children to that tab"?  I don't really feel like trying to find / catch / prevent this is worth it...

Related: would it be worth saying in the docs explicitly that a floating point exception will occur if you do this?  It might help future users who accidentally do this understand what is wrong, but it may be overkill

svenevs:
- An exception is always raised (rather than a floating point signal)
- Brief example in class level documentation

Fixes #303.  The approach here is mostly what is described by @Thorius [in this comment](https://github.com/wjakob/nanogui/issues/303#issuecomment-361759831), only overriding the `Widget::addChild(int, Widget*)` method instead.

Breakdown of how / why it works:

1. When constructing a widget with the parent specified, this is all that really happens:

    https://github.com/wjakob/nanogui/blob/f61c407ea4cb336aefcdd82202c6e09ff21b9e79/src/widget.cpp#L28-L29

    As such, when creating `mHeader` and `mContent`, we first explicitly orphan them by initializing with `nullptr`:

    ```cpp
    TabWidget::TabWidget(Widget* parent)
        : Widget(parent)
        , mHeader(new TabHeader(nullptr)) // create using nullptr, add children below
        , mContent(new StackedWidget(nullptr)) {
    ```

    and then deliberately call `Widget::addChild` in the constructor body

    ```cpp
    // since TabWidget::addChild is going to throw an exception to prevent
    // mis-use of this class, add the child directly
    Widget::addChild(childCount(), mHeader);
    Widget::addChild(childCount(), mContent);
    ```

    this means that `TabWidget::addChild` is not called, but theme setting / reference counting is still updated correctly.

2. An exception is always raised in `TabWidget::addChild`, signaling as early as possible that this should not be done (rather than deferring to `performLayout` as discussed in #303).  As a side-effect, if somebody does really need to modify `TabWidget` and add more children in some derived class, presumably they will see what is going on in the constructor and figure it out.

    A simple test is to apply this diff to `example1.cpp`, which will trigger the desired exception:

    ```diff
    --- a/src/example1.cpp
    +++ b/src/example1.cpp
    @@ -340,6 +340,8 @@ public:

             TabWidget* tabWidget = window->add<TabWidget>();

    +        new Label(tabWidget, "This causes an exception");
    +
             Widget* layer = tabWidget->createTab("Color Wheel");
             layer->setLayout(new GroupLayout());
    ```

I think this is the cleanest possible way of doing this, while still leaving the door open for people to derive `TabWidget` if they desire.  Attached for convenience are the rendered docs explaining how to use this:

tab_docs

### DIFF
--- a/include/nanogui/tabwidget.h
+++ b/include/nanogui/tabwidget.h
@@ -25,10 +25,55 @@ NAMESPACE_BEGIN(nanogui)
  *
  * \brief A wrapper around the widgets TabHeader and StackedWidget which hooks
  *        the two classes together.
+ *
+ * \rst
+ *
+ * .. warning::
+ *
+ *    Unlike other widgets, children may **not** be added *directly* to a
+ *    TabWidget.  For example, the following code will raise an exception:
+ *
+ *    .. code-block:: cpp
+ *
+ *       // `this` might be say a nanogui::Screen instance
+ *       Window *window = new Window(this, "Window Title");
+ *       TabWidget *tabWidget = window->add<TabWidget>();
+ *       // this label would be a direct child of tabWidget,
+ *       // which is forbidden, so an exception will be raised
+ *       new Label(tabWidget, "Some Label");
+ *
+ *    Instead, you are expected to be creating tabs and adding widgets to those.
+ *
+ *    .. code-block:: cpp
+ *
+ *       // `this` might be say a nanogui::Screen instance
+ *       Window *window = new Window(this, "Window Title");
+ *       TabWidget *tabWidget = window->add<TabWidget>();
+ *       // Create a tab first
+ *       auto *layer = tabWidget->createTab("Tab Name");
+ *       // Add children to the created tabs
+ *       layer->setLayout(new GroupLayout());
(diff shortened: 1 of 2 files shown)


wjakob:
Looks great, thank you!

PR #313, opened 2018-02-16
export Serializer, example5 on serialization

svenevs:
Supercedes #261.

> I suspect that all of the LTO-related issues are simply due to the fact that the class declaration of the serializer reads
> 
> ```cpp
> class Serializer {
> ```
> and not
>
> ```cpp
> class NANOGUI_EXPORT Serializer {
> ```

**WOW** I can't believe how easy that should have been.  This PR just adds the `NANOGUI_EXPORT`, some `Serializer` docs, and `example5.cpp`.  It does **not** fix serialization of Themes (#210) completely, as serializing themes is actually much more convoluted.  I'll add some thoughts on it on that thread.  This PR also omits the changes to `glutil.h` from the other one, I'll fix those at a later time (I found the docs bug, just need to fix it).

### DIFF
--- a/src/example5.cpp
+++ b/src/example5.cpp
@@ -1,0 +1,264 @@
+/*
+    src/example5.cpp -- example of how to use the same serializer to save the
+    state of many different widgets using the `push` and `pop` feature of
+    the Serializer class to create "namespaces" for each widget being saved.
+
+    NanoGUI was developed by Wenzel Jakob <wenzel.jakob@epfl.ch>.
+    The widget drawing code is based on the NanoVG demo application
+    by Mikko Mononen.
+
+    All rights reserved. Use of this source code is governed by a
+    BSD-style license that can be found in the LICENSE.txt file.
+*/
+
+// this define just makes it so when we include the example1.cpp file,
+// the `main` method is not included
+// tip: including source files is BAD PRACTICE!  this was done out of laziness,
+//      since example1 uses most widgets, we simply inherit that
+//      ExampleApplication and serialize that
+#define NANOGUI_EXAMPLE_5_BYPASS
+#include "example1.cpp"
+
+#include <nanogui/serializer/core.h>
+#include <nanogui/serializer/opengl.h> // unused in this example
+#include <nanogui/serializer/sparse.h> // unused in this example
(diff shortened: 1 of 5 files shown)


src/example5.cpp
```diff
@@ -0,0 +1,264 @@
...
+            nanogui::ref<ExtendedExampleApplication> app = new ExtendedExampleApplication();
+            app->drawAll();
+            app->setVisible(true);
+            nanogui::mainloop();
+        }
+
+        nanogui::shutdown();
+    }
+    catch (const std::runtime_error &e) {
+        std::string error_msg = std::string("Caught a fatal error: ") + std::string(e.what());
```
svenevs: To be honest, I don't even remember why I have this stuff in here.  Should the `try-catch` just be removed?

wjakob:
I've been a bit hesitant to merge this. Basically it seems very complex to me (the code of setting all the labels, and then the need of pushing/popping labels in the serializer to get the job done). The serializer feature was extensively used in the Instant Meshes project, where the main change was to call ``setId`` for various important widgets (e.g. https://github.com/wjakob/instant-meshes/blob/011fa44ab72037cbc16535090a63ead41963c1e5/src/viewer.cpp#L418). I basically worry that this example is not very representative of how the feature would be used in practice, which limits the educational value.

svenevs:
A fair assessment, TBH writing the example was the first time I ever used the `Serializer`.  I ran into issues without pushing / popping so I thought that was the intended purpose.

So, as a followup, what do you want to see in the example serialization?

1. The primary incentive for inheriting from Example 1 is it's a really good test case -- there are widgets that do not properly set / get.  I took a look at the `setId` stuff, and doing that would pretty much irradicate the need for the ridiculous `std::map` I made.  So something like

    ```cpp
    int idx = 0;
    for (auto *child : mChildren) child->setId("child_" + std::to_string(idx++));
    ```

    But I think what you're getting at is serializing **everything** is what makes the example so confusing.

2. With that in mind, are you looking instead for something that just serializes a few key widgets?  That would make things much clearer, but sacrifices the baked-in test-case-for-nearly-every-widget.

PR #317, opened 2018-02-23
Custom fonts

svenevs:
Replaces #282 with a much cleaner history.

**BEFORE MERGE**

- [x] Update the docs
- [x] Test windows build.
- [x] Place `popup` theme setting responsibility to parent (with docs indicating this).  So `ComboBox` and `PopupButton`.
- [x] Get rid of `FontWidget` and put that all in `Widget`.
- [x] ~~Allow support for `.otf` in addition to `.ttf`.~~ redacted since requires libfreetype.
- [x] Fix the [checkbox bindings](https://github.com/wjakob/nanogui/pull/317#discussion_r171773813)
- [x] Add [tooltip opacity theme override](https://github.com/wjakob/nanogui/pull/317#discussion_r171774399)
- [x] restructure obtuse `mFont` / `mFontExplicit` logic.
- [x] post-mortem / rewrite history

This PR is large, but solves once and for all the ability to embed custom fonts.  I will add a review highlighting the 2 problems.  I'm opening up the PR for review now, docs and Windows testing TBD after I finish the docs on the subsidiary repos (there's an automated icon font generation repo!).

1. `bin2c` generates to `${CMAKE_CURRENT_BINARY_DIR}/nanogui`.  Makes installation easier (`nanogui_resources.h` was not being installed), and keeps the `NANOGUI_EXTRA_INCS` clean.
2. ~~No font names are hard-coded in any implementation, **except** `Screen::tooltipFont`.~~ Changed, `Widget::mTooltipFont` defines this per-widget now.
3. `Theme` defines the default fonts.
4. ~~Any class that was drawing non-icon font now inherits from `FontWidget`.~~ It's all in `Widget` now.
5. Python bindings with custom **icon** fonts was a little contrived, but (a) I don't think many people will use it, it's there to allow projects that really cannot use CC-BY-SA 4.0 from Entypo+, (b) there are numerous safeguards in place.

For build system changes, see first commit message.


include/nanogui/widget.h
```diff
@@ -186,13 +186,6 @@ class NANOGUI_EXPORT Widget : public Object {
     const std::string &tooltip() const { return mTooltip; }
     void setTooltip(const std::string &tooltip) { mTooltip = tooltip; }
 
-    /// Return current font size. If not set the default of the current theme will be returned
-    int fontSize() const;
```
svenevs: Moving these to `FontWidget` introduces the only potential API breaking change.  If people have custom widgets that relied on these, they will need to inherit from `FontWidget` now instead.

Is there a way to make this clear for users who encounter this in the future?  Is there an alternative?

src/popup.cpp
```diff
@@ -20,6 +20,9 @@ NAMESPACE_BEGIN(nanogui)
 Popup::Popup(Widget *parent, Window *parentWindow)
     : Window(parent, ""), mParentWindow(parentWindow),
       mAnchorPos(Vector2i::Zero()), mAnchorHeight(30), mSide(Side::Right) {
+    // Make sure the theme propagates to future children
+    if (mParentWindow)
+        this->setTheme(mParentWindow->theme());
```
svenevs: I got really turned around with what to do here.  Take `PopupButton`, for example.  Neither `parent` nor `parentWindow` will ever be the actual `PopupButton` instance.  Therefore when `Widget::addChild` is called, the `Theme` setting gets messed up.

It seems pretty reasonable to assume that custom theme users will set them per window, and they can always call `setTheme` manually.  I thought about adding documentation, but didn't know what to say, and feel like there is a better solution here.

It technically only works in my example repo because I'm explicitly setting the `Theme` of a window.

svenevs:
Hmm. @wjakob what are your thoughts about making font loading a static action for `nanogui::Theme`?  I don't think it can be done in `nanogui::init()` because we need the `NVGContext`.  But

1. [Font memory cannot be freed](https://github.com/memononen/nanovg/issues/337).
2. Technically, custom themes currently re-load all fonts from the base class.

For example, with the following in `nanogui::Theme`:

```cpp
mFontNormal     = createFontMem(ctx, "sans", "Roboto-Regular.ttf");
mFontBold       = createFontMem(ctx, "sans-bold", "Roboto-Bold.ttf");
mFontMonoNormal = createFontMem(ctx, "mono", "RobotoMono-Regular.ttf");
mFontMonoBold   = createFontMem(ctx, "mono-bold", "RobotoMono-Bold.ttf");
mFontIcons = createFontMem(ctx, "icons", "entypo.ttf");
```

loads another font, all of them are re-loaded (memory leak).  A simple solution would be to introduce a static boolean

```cpp
static bool fonts_loaded = false;

Theme::Theme(NVGContext *ctx) {
    // ...
    if (!fonts_loaded) {
        // ... load fonts ...
        fonts_loaded = true;
    }
}
```

The same practice would be encouraged with people who want to load custom fonts in a derived class, but if people are just customizing colors / icons, they don't need to worry about this anymore.

As far as I can tell, `mFontNormal` etc are not needed.  We need to check that the return value is not `-1` (loading was successful), but the actual value of it is not used (at least be NanoGUI).  What you register the font name as (e.g., `"sans"`) is what is needed.

include/nanogui/widget.h
wjakob: I suggest keeping a single Widget class (i.e. no ``FontWidget``). In the bigger scheme of things, widgets are small enough memory-wise that we don't need to care about e.g. some extra string fields.

wjakob:
I agree, it's not worth the trouble. The thing that is important is that users can easily register new fonts and use them with all Widgets, which this PR accomplishes nicely.

CMakeLists.txt
```diff
@@ -253,31 +253,150 @@ endif()
...
+# If all checks are successful, then the icon font will be included with
+# NANOGUI_EXTRA_RESOURCES, the header file will be copied to
+# ${CMAKE_CURRENT_BINARY_DIR}/nanogui (which is included with NANOGUI_EXTRA_INCS),
+# and if python is being built, the configurations for python/main.cpp.in will
+# be coordinated to bind the font for python.
+if (NANOGUI_EXTRA_ICON_RESOURCES)
+  foreach(icon_font ${NANOGUI_EXTRA_ICON_RESOURCES})
+    # Make sure the font is a true type font
+    get_filename_component(icon_font_ext "${icon_font}" EXT)
+    if (NOT "${icon_font_ext}" STREQUAL ".ttf")
```
wjakob: what about OTF?

include/nanogui/widget.h
svenevs: That can be done.  The reason `FontWidget` became a thing was so I could force three parameters:

```cpp
FontWidget(Widget *parent, const std::string &font, bool fontDefaultIsBold)
```

1. If `font = ""`, then `setDefaultFont` takes over and sets it based on the theme. 
2. `fontDefaultIsBold` determines what to query from `Theme` (added virtual methods `defaultFont()` and `defaultBoldFont()`.
3. When `setTheme` is called, a boolean flag `mFontExplicit` signals whether or not the user explicitly requested a font e.g. for the `Label` (also is set to `true` when `setFont` is called).

So when putting this all in `Widget`, there are two reasonable options:

1. Add those parameters to the `Widget` constructor with default values (API compatibility).  The implication is that **every** widget gets a font face set (perhaps that is ideal).
2. Require that sub-classes who want font faces first set `mFont` and `mFontDefaultIsBold` in their derived constructor, and call `setDefaultFont()` afterward.

Somewhat related, do you care if I force-push here?  Since the changes are so many, I tried to make the commits mergeable rather than squash so that in the future when people are looking at the history, they don't get overwhelmed.

CMakeLists.txt
svenevs: Good point!

svenevs: @wjakob how much do you care about `.otf`?  [NanoVG does not support them](https://github.com/memononen/nanovg/issues/37).  The issue describes an alternative, but I think the changes required to support it would be much further reaching than just using [fontstash](https://github.com/memononen/fontstash) to load fonts.  I think we'd have to update **all** code drawing fonts to use that over NanoVG...

src/label.cpp
```diff
@@ -17,27 +17,29 @@
...
-    }
 }
 
 Vector2i Label::preferredSize(NVGcontext *ctx) const {
     if (mCaption == "")
         return Vector2i::Zero();
-    nvgFontFace(ctx, mFont.c_str());
-    nvgFontSize(ctx, fontSize());
+
+    float fontSize = Widget::fontSize(mTheme->mStandardFontSize);
```
svenevs: First occurrence of many, `Widget::fontSize` has changed meaning.  Supply a default value (`mTheme->{mStandardFontSize,mTextBoxFontSize}` etc).  If `Widget::mFontSize` has been set explicitly via `setFontSize`, then `mFontSize` is returned.  Otherwise, the default value specified by the caller is returned.

All font sizes have been changed to floats, since the `nvgFontSize` method parameter is a `float` (e.g., `Theme` font sizes are all floats now but with same values).

CMakeLists.txt
```diff
@@ -253,31 +253,192 @@ endif()
...
+  foreach (font ${NANOGUI_EXTRA_RESOURCES})
+    get_filename_component(font_ext "${font}" EXT)
+    if (NOT "${font_ext}" STREQUAL ".ttf" AND NOT "${font_ext}" STREQUAL ".otf")
+      message(FATAL_ERROR "NanoGUI can only embed true type and open type fonts.  [${font}] does not end with '.ttf' or '.otf'.")
+    endif()
+
+    if ("${font_ext}" STREQUAL ".otf")
+      if (NOT FREETYPE_FOUND)
+        message(
+          WARNING "[${font}] is an Open Type Font (.otf), but Freetype was not "
```
svenevs: I wasn't sure whether to make this a `WARNING` or a `FATAL_ERROR`.  I chose `WARNING` because it appears freetype may be cumbersome on Windows.  I was planning on just incorporating in the docs how to accommodate this.  The idea is basically that `-1` will be returned when *loading* the font, so at that point you can either fail out like `Theme` does with an `std::runtime_error`, or you could choose to fallback.

svenevs:
@wjakob I believe this is *finally* ready.  I don't have a bootable windows right now (I selfishly tried to take my boot cycle back to linux as default....&$#*&$#*), but the Appveyor tests at least build both here and [for actually embedding the fonts](https://ci.appveyor.com/project/svenevs/nanogui-customization-demo), but that doesn't test runtime.

See comments above from first review on highlighted cases that may need to be addressed.  In addition to making sure font and theme setting propagate correctly with `setTheme`, **there are unrelated changes present here**:

1. [Enable clipboard manipulations in python](https://github.com/wjakob/nanogui/pull/317/commits/b8dd5e67e3072e05541e2fa12bf5f35ef3156dc7).  This was really useful for the [theme builder](https://github.com/svenevs/nanogui-customization-demo#theme-builder).  I bound it to `Screen` because I couldn't bind it in the `glfw` module.  It would compile, but not exist (probably because we'd need to bind `GLFWwindow` which I didn't really want to do...).
2. `ColorWheel` previously always returned `1.0` for alpha, [that is now changed](https://github.com/wjakob/nanogui/pull/317/commits/55467f3577d4a08f4d188cf00d52a396543278d5).  This caused problems when the `ColorPicker` value is used directly in a callback (meaning `alpha` cannot be modified intermediately).  I don't see anything wrong with it, but maybe there should also be a corresponding `setAlpha` method?  I didn't need it for the Theme Builder application though.
3. Docs: I added a section on [setting up a `git submodule`](http://fake-nanogui.readthedocs.io/en/custom_fonts/compilation.html#configuring-a-git-submodule) since that's come up in a couple of issues.  I think it is descriptive enough for users new to submodules, but perhaps it could be more concise.

Where docs are concerned, as opposed to last time with the Entypo stuff, everything has been fully tested and working.  However, I am not very good at keeping things concise.  The docs updates are as follows (if you want to change it, you should be able to commit directly on this branch):

1. [Linked the compilable examples for customization](http://fake-nanogui.readthedocs.io/en/custom_fonts/examples.html#example-customization)
2. Compilation has two new sections: [Including Custom Fonts](http://fake-nanogui.readthedocs.io/en/custom_fonts/compilation.html#including-custom-fonts) and [Including Custom Icon Fonts](http://fake-nanogui.readthedocs.io/en/custom_fonts/compilation.html#including-custom-icon-fonts)
3. The [Usage](http://fake-nanogui.readthedocs.io/en/custom_fonts/usage.html) page now details exactly how fonts / icons are used, as well as how to modify defaults via custom theme classes.

svenevs:
Ugh sorry, there is one more thing that needs to be resolved: [what to do about `CheckBox` bindings](https://github.com/wjakob/nanogui/pull/317#discussion_r171773813)

src/screen.cpp
```diff
@@ -443,10 +443,10 @@ void Screen::drawWidgets() {
                 h = (bounds[2] - bounds[0]) / 2;
             }
             nvgGlobalAlpha(mNVGContext,
-                           std::min(1.0, 2 * (elapsed - 0.5f)) * 0.8);
+                           std::min(1.05, 2 * (elapsed - 0.5f)) * 0.8);
```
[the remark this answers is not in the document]
wjakob: Yes, you could add a theme variable for this. (mTooltipOpacity?)

wjakob:
FYI: This PR doesn't compile out of the box on my machine (OSX):

```bash
$ mkdir build
$ cmake -GNinja ..
[..] omitted [..]
$ ninja 
[64/10/1] Building CXX object CMakeFiles/nanogui-python-obj.dir/python/main.cpp.o
FAILED: CMakeFiles/nanogui-python-obj.dir/python/main.cpp.o 
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++   -DCORO_SJLJ -DGLAD_GLAPI_EXPORT -DNANOGUI_PYTHON -DNANOGUI_SHARED -DNVG_SHARED -I../ext/eigen -I../ext/glfw/include -I../ext/nanovg/src -I../include -I. -I../ext/coro -I../ext/pybind11/include -I/Users/wjakob/anaconda/include/python3.6m -fvisibility=hidden  -Wall -Wextra -std=c++14 -flto=thin -stdlib=libc++ -O3 -DNDEBUG -fPIC   -fobjc-arc -Os -MD -MT CMakeFiles/nanogui-python-obj.dir/python/main.cpp.o -MF CMakeFiles/nanogui-python-obj.dir/python/main.cpp.o.d -o CMakeFiles/nanogui-python-obj.dir/python/main.cpp.o -c /Users/wjakob/projects/nanogui/python/main.cpp
/Users/wjakob/projects/nanogui/python/main.cpp:7:10: fatal error: 'nanogui/resources.h' file not found
#include <nanogui/resources.h>
         ^~~~~~~~~~~~~~~~~~~~~
1 error generated.
[64/1/10] Building CXX object CMakeFiles/nanogui-python-obj.dir/python/button.cpp.o
ninja: build stopped: subcommand failed.

```

svenevs:
Hmmm. Ok I'll check `ninja` on OSX later today.  It should have generated `${CMAKE_CURRENT_BINARY_DIR}/nanogui/resources.{h,cpp}`, so that doing `include_directories(${CMAKE_CURRENT_BINARY_DIR})` finds `<nanogui/resources.h>`.

I'll add the tooltip theme bypass as well.

Any thoughts on the CheckBox python binding code?

include/nanogui/widget.h
```diff
@@ -334,7 +413,66 @@ class NANOGUI_EXPORT Widget : public Object {
...
+
+    /// The current font face being used to draw text (if this Widget displays text).
+    std::string mFont;
+    /**
+     * When the font is specified either via the constructor or \ref setFont,
+     * this is set to ``true`` to indicate that \ref setTheme should not
+     * overwrite the user-specified default.  Otherwise, when \ref setTheme is
+     * called, the theme's default window font will be used.
+     */
+    bool mFontExplicit = false;
```
wjakob: Hi @svenevs,

the amount of storage and logic dedicated to keep track of fonts in the ``Widget`` class has me somewhat concerned. I feel like it's best if there is one more iteration that simplifies this as much as possible.

For starters, why do we need ``mFontExplicit``? I assume that the font will just be an empty string initially. ``setFont`` could be implemented so that it only replaces the font string in ``mFont`` when it is equal to the empty string (or the corresponding font string from the previous theme, which I suppose is quite rare usage and shouldn't drive complexity too much here).

include/nanogui/widget.h
```diff
@@ -334,7 +413,66 @@ class NANOGUI_EXPORT Widget : public Object {
...
+    /**
+     * There are two default fonts in NanoGUI: ``"sans"`` and ``"sans-bold"``.
+     * When this variable is ``true``, the bold font will be used.  When
+     * ``false``, the regular font face will be used.  Subclasses are
+     * responsible for specifying this behavior when calling the parent class
+     * constructor (\ref Widget::Widget).  For example, \ref nanogui::Label will
+     * always set this to ``false``, whereas \ref nanogui::Button will always
+     * set this to ``true``.
+     */
+    bool mFontDefaultIsBold;
```
wjakob: The same here. I don't think it's a good tradeoff to include a top-level field just to indicate whether a widget wants a bold font by default. You could instead have ``setDefaultFonts`` be a virtual method that is overridden by (few) widgets.

python/button.cpp
```diff
@@ -46,27 +48,49 @@ void register_button(py::module &m) {
...
+        .def("setCallback", &ComboBox::setCallback, D(ComboBox, setCallback))
+        .def("selectedIndex", &ComboBox::selectedIndex, D(ComboBox, selectedIndex))
+        .def("setSelectedIndex", &ComboBox::setSelectedIndex, D(ComboBox, setSelectedIndex))
+        .def("setItems", (void(ComboBox::*)(const std::vector<std::string>&)) &ComboBox::setItems, D(ComboBox, setItems))
+        .def("setItems", (void(ComboBox::*)(const std::vector<std::string>&,
+                          const std::vector<std::string>&)) &ComboBox::setItems/*, D(ComboBox, setItems, 2)*/)
+        .def("items", &ComboBox::items, D(ComboBox, items))
+        .def("itemsShort", &ComboBox::itemsShort, D(ComboBox, itemsShort));
+
+    std::function<void(bool)> checkbox_default_cb = [](bool) { /* does nothing */ };
```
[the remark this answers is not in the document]
wjakob: Slightly counter-intuitive but not a Pybind11 bug. Pybind11 by default only handles ``std::functions`` when doing auto-conversion of functions. 

In this case, a much more sensible thing would be to write ``py::arg("callback") = py::none()`` in my opinion.

src/label.cpp
wjakob: Sounds good!

wjakob:
Wowowow -- I love the theme builder!! We should be sure to reference it extensively in the NanoGUI docs.

wjakob:
This is shaping up to be an amazing PR!! Thank you so much for spending so much time on this.

I added a few specific code comments above. The documentation changes look great -- this will be very valuable to others (and myself, for reference :)). 

Really the only (small) issue that I see is the large number of font-related extra fields in the ``Widget`` class, which should be cut down. Other than that, this is good to go!

wjakob:
By the way: irrelevant to this PR, but I thought you might be interested.

I've been (slowly) porting over all patches here to a forked NanoGUGI repository. It is based on a new vector library that will be the foundation of Mitsuba 2 and also has slightly different coding conventions to make for a more seamless Python <-> C++ transition (https://github.com/mitsuba-renderer/nanogui). This version now also supports GLES and WebAssembly code generation, which is loads of fun: http://mitsuba-renderer.org/nanogui.html (works in Chrome and recent Firefox, haven't tried other browsers)

src/screen.cpp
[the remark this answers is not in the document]
svenevs: That can be done!  The only snag is that the side-by-side stuff in the repo with theme builder got pretty obtuse, `make_compare_window`(https://github.com/svenevs/nanogui-customization-demo/blob/77a95c38501f9e50f0ac7770e4991df632130bff/cpp/make_compare_window.hpp#L50) is beastly.

1. Which widgets would you like to show up in the side-by-side comparison window?

    The linked function has all of that because each of those is needed for testing the theme changes.  But having all of that code in the example would distract from the purpose.

2. I switched from the light theme to a not-so-great "solarized light"-esque theme in that repo.  You want this example to have a more traditional light theme (like what's in the pictures above), right?

3. Can I include the `json` comment with a link to the `theme-builder`?

include/nanogui/widget.h
svenevs: Yes, there has to be a better way to approach this.  The tension is really just that there is no certainty about when `setTheme` is called (e.g. before or after `new Label(parent, "text", "sans-bold")`), so these really only exist to ensure that `setTheme` does not overwrite user-specified font values.

So it sounds like the goal here would be to merge `mFontExplicit` to be implicit when `mFont == ""`?.  So (ignoring bold stuff), `font()` could change to be

```cpp
if (mFont != "")
    return mFont;
return mTheme->defaultFont();
```

That alleviates the concerns about theme setting overwriting user-specified values.  I don't see a solution for default vs bold other than a "magic" value.  E.g.

```cpp
if (mFont != "") {
    if (mFont == "THEME_BOLD")
        return mTheme->defaultBoldFont();
    else
        return mFont;
}
return mTheme->defaultFont();
```

I was trying to avoid stuff like this because now we're forcing a virtual method into the rendering loop for all widgets using fonts, when most users likely won't be using custom themes in the first place.

I will need to think more carefully about how to solve this.

svenevs:
> This version now also supports GLES and WebAssembly code generation, which is loads of fun: http://mitsuba-renderer.org/nanogui.html

WOAH!  I'm gonna have to take a closer look at that soon, that's really cool!  Maybe down the road I will figure out how to port the theme builder so that people can just use their browser :upside_down_face:

include/nanogui/widget.h
svenevs: @wjakob I've thought about this more, and I don't think avoiding virtual methods can be done here.  Before implementing, do you think this is a good approach?

1. `mFont` is the only `Widget` variable.  It remains `private` necessitating users to call `font()`.
2. `Widget::font()` implementation becomes

    ```cpp
    if (mFont != "")
        return mFont;
    return defaultFont();
    ```
3. Remove font related methods in `Theme` in favor of three instance variables (also makes customization even easier): `Theme::mDefaultFont`, `Theme::mDefaultBoldFont`, and `Theme::mDefaultIconFont`.
4. Introduce a virtual `Widget::defaultFont()`, which subclasses override if they want to specify a default font of bold.

    ```cpp
    virtual std::string defaultFont() {
        if (mTheme)
            return mTheme->mDefaultFont;
        return "";
    }

    The only time the empty string will be returned is if somebody explicitly created an orphan widget.  But as far as I understand, it is assumed that `mTheme` will exist when rendering takes place.  So do we even want the check?

python/button.cpp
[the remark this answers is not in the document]
svenevs: PyBind is so impressive!  The following all work now:

```pycon
>>> import nanogui
>>> nanogui.init()
>>> s = nanogui.Screen([400, 400], "Test")
>>> w = nanogui.Window(s, "Some Window")
>>> w.setLayout(nanogui.GroupLayout())
>>> nanogui.CheckBox(w, "First")
<nanogui.CheckBox object at 0x7f20f7eac810>
>>> nanogui.CheckBox(w, "Second", lambda x: print("2: {}".format(x)))
<nanogui.CheckBox object at 0x7f20f7eac990>
>>> nanogui.CheckBox(w, "Third", callback=lambda x: print("3: {}".format(x)))
<nanogui.CheckBox object at 0x7f20f7eac810>
>>> s.setVisible(True)
>>> s.performLayout()
>>> nanogui.mainloop()
```

svenevs:
@wjakob 

1. FYI I had to force-push here (hence all the new commits), I needed to rebase on to master to get the CI to run again (PopupButton had a conflict).  Everything here is what you reviewed previously except for the last three commits.

2. RE: Ninja.  I'm not sure what to do here.  This is a CMake bug.  To add to the confusion, both locally and on Travis (added a Ninja specific CI for OSX) [`ninja || ninja` will work](https://travis-ci.org/wjakob/nanogui/jobs/371905458#L1186)?!

    I reached out to the CMake mailing list to ask for advice on how to bypass this, it's some kind of tension between Ninja wanting relative paths and using `include_directories(${CMAKE_CURRENT_BINARY_DIR})`.

    I will post back if I get a response, but if no solution exists what do you want to do?  The reason for generating into a `nanogui/` sub-directory was for two reasons: consistency (`#include <nanogui/resources.h>`) and installability.  If we keep `nanogui_resources.h`, then that file has to install directly into `/usr/include` (for example), rather than `/usr/include/nanogui`.

svenevs:
Holy damn that was so painful.  @petterarvidsson you may be interested in this since you wrote the (magical) `resources/bin2c.cmake`.

So the real problem here is not to do with `ninja` and path issues, but `add_custom_command` and order of operations.  Though it is unclear why this **previously worked** generating `nanogui_resources.{h,cpp}`, the argument is that it should not have been.

1. CMake configuration is run, this produces the build system e.g. a `Makefile` or `ninja.build`.
2. For most generators, it appears that `PRE_BUILD` in `add_custom_command` will work as expected.  **However**, it's actually only a guarantee that it will truly be `PRE_BUILD` [for Visual Studio 8 and higher](https://cmake.org/cmake/help/latest/command/add_custom_command.html#build-events).
    - Locally on Linux and Ninja it worked fine, but for OSX it is becoming `PRE_LINK` as the docs promise, which is where the `#include <nanogui/resources.h>` failure comes from --- it doesn't exist yet.

There is technically a way to cheat this through an `add_custom_target(nanogui-resources-dependency)` and some really arcane hacking.  But then I realized we can actually just run `bin2c.cmake` at configure time, fixing this once and for all.

@wjakob I added two CI to Travis.

1. OSX // Ninja, works!!!
2. Linux // Ninja: total nightmare.  I tried every `g++` possible (up to and including `g++-7`), and many `clang`, with varying results of failure caused by outrageously underwhelming worker setups.  Currently on this PR the Linux // Clang // Ninja works **sometimes**, but `ninja` typically fails for unknown reasons (eventually the worker just kills `clang`).
    - Option 1: delete the Clang Linux CI (but Clang is Clang!)
    - Option 2: keep the Clang Linux CI, but use Makefiles.  That [reliably works](https://travis-ci.org/svenevs/nanogui/jobs/372337154).

Sorry for so many force pushes / so much noise around this.  This was one of the most frustrating problems I've solved in a while, everything that worked locally would fail on CI!

CMakeLists.txt
```diff
@@ -372,38 +373,23 @@ if (NANOGUI_EXTRA_RESOURCES)
...
-# Run bin2c on resource files
-add_custom_command(
-  OUTPUT nanogui/resources.cpp nanogui/resources.h
-  COMMAND ${CMAKE_COMMAND} ARGS ${bin2c_cmdline}
-  DEPENDS ${resources}
-  COMMENT "Running bin2c"
-  PRE_BUILD VERBATIM)
+# All font resources are now stored in `resources` list.  Execute bin2c at
+# configure time, see top of resources/bin2c.cmake for more information.
+set(BIN2C_OUTPUT_C "${NANOGUI_GENERATED_RESOURCE_DIR}/src/resources.cpp")
```
wjakob: Very minor: since these are now part of the NanoGUI CMake scope, I propose prefixing with ``NANOGUI_``. These variables sometimes show up in visual CMake frontends, so it's nice to immediately see where they come from.

wjakob:
Hi Stephen,

thanks for looking into this. I never even realized that there was a problem with bin2c :), and this looks like a good fix. It's interesting that Ninja fails on Travis. Do you think that this is a fault with Ninja itself, rather than something associated with this project? In any case, switching to Makefiles sounds like a perfectly good workaround.

Best,
Wenzel

include/nanogui/widget.h
wjakob: +1 for this approach by the way.

CMakeLists.txt
svenevs: WILCO, I'll also prefix `NanoGUI:` in the `message(STATUS`.  The more fonts requested the longer it takes, making nice to say what step is going on ;)

svenevs:
TBH I have no idea.  I lean toward Ninja, I'll see if I can get a newer version easily enough on Travis.  The linux one is `v1.3.4`, whereas with `brew` on OSX its `v1.8.2`.

In the final checkbox item (reduce font logic complexity) I am currently stuck on something that did not happen before, something is up with the `TabWidget` boundaries :sigh:.  That and a small rewrite of the docs (fonts revisited are much cleaner!) should be available soon.

CMakeLists.txt
svenevs: So the output looks like this:

```console
-- NanoGUI: generating font resources.
--   Processing: Roboto-Bold.ttf
--   Processing: Roboto-Regular.ttf
--   Processing: RobotoMono-Bold.ttf
--   Processing: RobotoMono-Regular.ttf
--   Processing: entypo.ttf
```

I like having the "Processing" part in there because the step does take a little bit now.  If you would rather not have it just let me know!

wjakob:
Hi Stephen,

is there anything that still needs to be done here? Otherwise please let me know and I'll merge it.

Best,
Wenzel

svenevs:
Hey sorry this kind of fell off my radar.  This is not merge ready.

I've got some local changes that reworked the `mFont` stuff, removing all of the `mFontExplicit` parts. That shouldn't take too long, but requires a small rewrite of the docs as it changed how theme bypassing works (virtual functions => member strings, makes it easier to override).

The tab widget problem actually happens on master, I think it may be a Wayland problem. I couldn't fix it, but I'll at least open a separate issue so it is documented.

I'll ping you when this is ready for merging, and we can discuss briefly how the history should be rewritten.

wjakob:
Okay, that sounds great!

svenevs:
@wjakob Ok I think this is good!

Post-mortem is four parts: code change, docs updates, potential change, and final TODO (rewrite history).

## Code

(What has changed since last review)

- `Widget::mFontExplicit` and the other two boolean flags are gone
    - The meat is in [`c052cb3`](https://github.com/wjakob/nanogui/pull/317/commits/c052cb34585c0e0d03969961abb2b06cf359c800)
    - Empty string `""` is used to indicate no user specified font (either via construction, or via `Widget::setFont`).
    - Theme `virtual` override methods are now just public `std::string mDefaultFont*` members.
        - Makes it so that users don't have to sub-class `nanogui::Theme` if they really don't want to
- Python bindings updated in suit.
- Added docs html and linkcheck builds to `.travis.yml`
    - The hacky `sed` and `$COLON` shenanigans will go away with exhale `v1.0.0` (in progress), I default doxygen to silent on ReadTheDocs and will do the same on Travis.  Travis is giving me a lot of grief with exhale right now.

## Documentation

Last 6 commits are [related to this ticket with RTD](https://github.com/rtfd/readthedocs.org/issues/4227).  I think we should try and merge this PR independently (I'll remove `.readthedocs.yml` since I don't know if it works).

1. Please build the docs locally to check `compilation.rst`.  Note that you need to use a virtual environment if you don't want your sphinx to downgrade.

    ```console
    $ cd docs/
    # it *must* be `venv` (this is documented in compilation.rst)
    $ virtualenv venv
    $ source venv/bin/activate
    $ pip install -r requirements.txt
    $ make html
    ```

    It's pinned at 1.6.7 because breathe is getting more and more broken.  I am still deciding whether I want to help fix it or just ditch it ;)

2. The pages of interest are updates to `usage` and `compilation`.  You have already seen the `compilation` additions, the `usage` customization sections are what i'm concerned about.  If you think it reads right then great!  But I've rewritten this document a few too many times and probably am missing some errors.

## Potential Changes

1. Instead of configuring directly `python/main.cpp.in` => `python/main.cpp`, would it be better to configure to `${CMAKE_CURRENT_BINARY_DIR}/generated_resources/python/main.cpp`?  Currently `generated_resources` has an `include/nanogui/resources.h` (and potentially custom icon font headers) and `src/resources.cpp`.  Generating here seems more consistent.

2. When you build locally, do you find the fact that running `bin2c` at configure time is annoying / too slow?  I can try and attempt a build-time solution with an artificial `nanogui-resources` target again, but if you don't care that much then cool :)

## Final TODO

Squashing this into one commit is probably a bad idea because so many files are affected.  I would like to rewrite the history here into the following categories, but doing this will mean that there will be a broken commit.  I don't know how I feel about this.

1. CMake build changes (builds will work)
2. CI updates (travis // Ninja business)
3. C++ changes, minus `include/nanogui/python.h` (python bindings break)
4. Python changes (build works again)
5. Docs updates to the new world order

I want to do it like this because it will be a lot easier for me / the whole reason to rewrite it is keep it manageable if somebody needs to bisect one of these one day.  If (3) is the only commit that does not build, but the commit message explicitly states that python is broken in this commit, is this an OK thing to do?

include/nanogui/theme.h
```diff
@@ -31,50 +31,83 @@ class NANOGUI_EXPORT Theme : public Object {
...
-     * setting both \ref Widget::mFont and \ref Widget::mTooltipFont.
-     */
-    static std::string globalDefaultFont() { return "sans"; }
-    /// Similar to \ref globalDefaultFont, only ``"sans-bold"`` is used.
-    static std::string globalDefaultBoldFont() { return "sans-bold"; }
-    /// Similar to \ref globalDefaultFont, only for icon fonts.
-    static std::string globalDefaultIconFont() { return "icons"; }
+     * \sa The *implementation* of \ref nanogui::Widget::font.
+     */
+    struct GlobalDefaultFonts {
```
svenevs: Oh...and this is the most ugly thing I've done in a while.  C++11 alternative would be `static constexpr auto Normal = "sans"` which I like very much, but I couldn't figure out how to bind that with pybind (I think because `constexpr` has no memory).  Is there a more modern alternative?

svenevs:
To anybody watching this, it would be very helpful if you can read the new usage documentation explaining how fonts are used / can be customized.  They are temporarily hosted here: https://fake-nanogui.readthedocs.io/en/custom_fonts/usage.html#how-fonts-are-used

If something does not make sense please tell me so I can fix it, otherwise this PR will be getting merged very soon.  Thanks!

CMakeLists.txt
```diff
@@ -356,10 +356,11 @@ if (NANOGUI_EXTRA_ICON_RESOURCES)
   endforeach()
 endif()
 
-# Run simple cmake converter to put font files into the data segment
-
-# Glob up resource files
-file(GLOB resources "${CMAKE_CURRENT_SOURCE_DIR}/resources/*.ttf")
+# Begin building the list of fonts to process with bin2c
+# These fonts are vendored with NanoGUI, but additional fonts may be added.
+foreach(font entypo.ttf Roboto-Bold.ttf RobotoMono-Bold.ttf RobotoMono-Regular.ttf Roboto-Regular.ttf)
```
svenevs: FWIW in the previous attempt at fixing whatever somehow broke, there were two things I missed / decided needed to fix

1. `nanogui-resources` target was not an explicit dependency of `nanogui-python-obj` (occasional failure on missing `#include <nanogui/resources.h>`.
2. I don't know for certain, but I think another issue was using `file(GLOB ...)`.  We know exactly which fonts we are embedding, and since it is a `DEPENDS` we want those to be explicit (different generators do different things with `GLOB`, CMake maintainers definitively encourage no `GLOB` unless it is absolutely necessary).

So this finally all works again **at build time**, [even on Ninja // OSX](https://travis-ci.org/wjakob/nanogui/jobs/393922647).

Maintenance note: this is mostly the same as the original `bin2c` layout, but I've kept the arguments as `NANOGUI_OUTPUT_H` etc rather than `OUTPUT_H`.  It seems more appropriate (even though it is not a configure time thing anymore), because what we do to process these files (adding them to the font map etc) makes it fairly inflexible in terms of other projects re-using this script.

svenevs:
Hi @wjakob,

I believe this is ready, I have fixed all potential issues mentioned in https://github.com/wjakob/nanogui/pull/317#issuecomment-396460231 .  In rewriting the history, I have revisited every change made here, and think this is as good as it is going to get.  I am happy to make additional changes, but I hereby decree my own stamp of "i'm done with this".  The changes basically amount to:

- Things that needed to change for custom fonts
- Things that needed to change for the `theme_builder` application to work (testing ground for theme setting)
- Associated continuous integration and documentation
- Some small misc commits fixing things that I found along the way.

I believe there is only one potential undesirable, and it is the last commit fixing the license year.  Would you prefer to use a version range instead?  I don't know when NanoGUI came about, but say it was 2014, then it would be `Copyright (c) 2014-2018 Wenzel Jakob <wenzel.jakob@epfl.ch>, All rights reserved` rather than the current `2018`.

The proposed history rewrite can be viewed [on a fake PR on my fork](https://github.com/svenevs/nanogui/pull/1).

The diff between that branch and this one is:

```diff
--- a/src/imageview.cpp
+++ b/src/imageview.cpp
@@ -67,10 +67,8 @@ namespace {
 }
 
 ImageView::ImageView(Widget* parent, GLuint imageID)
-    : Widget(parent, ""), mImageID(imageID), mScale(1.0f),
-      mOffset(Vector2f::Zero()), mFixedScale(false), mFixedOffset(false),
-      mPixelInfoCallback(nullptr) {
-
+    : Widget(parent), mImageID(imageID), mScale(1.0f), mOffset(Vector2f::Zero()),
+    mFixedScale(false), mFixedOffset(false), mPixelInfoCallback(nullptr) {
     updateImageParameters();
     mShader.init("ImageViewShader", defaultImageViewVertexShader,
                  defaultImageViewFragmentShader);
diff --git a/src/widget.cpp b/src/widget.cpp
index af0534f..41888f4 100644
--- a/src/widget.cpp
+++ b/src/widget.cpp
@@ -47,7 +47,7 @@ void Widget::setTheme(Theme *theme) {
 }
 
 float Widget::fontSize(const float &defaultFontSize) const {
-    if (defaultFontSize <= 1.0f)
+    if (defaultFontSize < 1.0f)
         throw std::runtime_error("Widget::fontSize: parameter must be greater than or equal to 1.0f.");
     return mFontSize < 0 ? defaultFontSize : mFontSize;
 }
@@ -304,6 +304,7 @@ bool Widget::load(Serializer &s) {
     if (!s.get("focused", mFocused)) return false;
     if (!s.get("tooltip", mTooltip)) return false;
     if (!s.get("cursor", mCursor)) return false;
+    if (!s.get("fontSize", mFontSize)) return false;
     if (!s.get("font", mFont)) return false;
     if (!s.get("tooltipFont", mTooltipFont)) return false;
     if (!s.get("iconFont", mIconFont)) return false;
```

1. `src/imageview.cpp`: no need to explicitly call `Widget(parent, "")` since default parameter for `font` is `""`.
2. `src/widget.cpp`: allow `defaultFontSize == 1.0f`, revive accidentally deleted serialization of `mFontSize`.

**Please let me know if I have a green light to force push those commits here** and merge this in.  This history rewriting business is good for future users / squashing this into one commit would be sad because it will make for an impossible diff to read.  _But I do not have the stamina to rewrite the history on this PR again_, it's rather painful.

svenevs:
Ok i went ahead and cherry-picked those commits, I think they present a very nice / clean history ready to be placed on master (rather than a merge or squash).  The commits all build independently and are mostly logical / will hopefully be easy to bisect if needed in the future.

1. There is one extra commit in there `[docs] more explicit Font Size documentation` which does exactly what it states.
2. I left `update license year` as the last commit so that if you want to switch to version ranges ( https://github.com/wjakob/nanogui/pull/317#issuecomment-399196230 ) that can be done with a force push easily.

P.S. Definitely not trying to rush anything if you haven't had time to review, I know you're very busy! Just figured the state i left this PR in two weeks ago makes it harder for review.

svenevs:
I'm closing this for this repository.  When future fork becomes live this can become multiple (**much smaller** / bite sized PRs).  The changes here help fix some nuanced bugs related to theme updates, but none of this is mission critical -- the things being fixed related to theme stuff are largely when you `setTheme()` on something after initial creation.

PR #321, opened 2018-03-08
Fix TDM-GCC build

rymut:
- Replace _stdcall with __stdcall keyword.
- Included required header for use of alloca function.
- Added dynamic loading mechanism for MonitorFromWindow function.

### DIFF
--- a/ext/glad/include/glad/glad.h
+++ b/ext/glad/include/glad/glad.h
@@ -27,7 +27,7 @@
 
 #if !defined(APIENTRY)
 #  if defined(_WIN32)
-#    define APIENTRY _stdcall
+#    define APIENTRY __stdcall
 #  else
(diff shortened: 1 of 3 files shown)


wjakob:
@rymut : These changes look good, thank you! Would you mind escalating the _stdcall vs __stdcall issue to the GLAD project? We're likely to reintroduce this issue in the future as the ``glad.c`` file is copied from this project.

rymut:
I do not mind doing some additional work, but please specify which glad project did you mean - there a lot of them. But most of them at least to my knowledge use __stdcall.

wjakob:
Never mind -- I just checked, and the official GLAD (https://github.com/Dav1dde/glad) actually doesn't generate this ``_stdcall`` annotation.

PR #324, opened 2018-03-18
accommodate CMake policy CMP0048

svenevs:
When parent projects declare `project(... VERSION X.Y.Z)` this warning surfaces, same fix as [pybind11/#567](https://github.com/pybind/pybind11/issues/567).


wjakob:
thanks!

Issue #326, opened 2018-04-04
Bug: Disposing a window with an open popup from a popup button segfaults


crazycarl:
I also have a working fix.

To produce the bug, create a window with a popup widget. Open the popup and then dispose the window. If you dispose the window while that popup is open, it will segfault while running the function `Popup::refreshRelativePlacement()`. It tries to access the pointer `mParentWindow`, which points to a window that has been disposed.

My fix was to remove the popup from the parent when the button is deconstructed.
```c
PopupButton::~PopupButton() {
    mPopup->parent()->removeChild(mPopup);
}
```
And I did a quick check with Valgrind to make sure it doesn't do anything bad with memory management. The memory management for this library is well designed, so no issues showed up.

crazycarl:
Looks like I jumped the gun a little. If the parent window is deleted before the PopupButton, this can cause an error where the reference count to the Window goes below zero.

I'm trying to find a better solution to this problem now.

crazycarl:
Okay, so after toying around a bit more and sleeping on the problem, I found that this fixes the issue without creating that reference count issue.

This just makes it so there's no reason to call `Popup::refreshRelativePlacement()` once the button itself is deleted.
```c
PopupButton::~PopupButton() {
    mPopup->setVisible(false);
}
```
Again, Valgrind reports no memory issues with this solution.

svenevs:
This seems like a good solution, I'm sure a PR would be welcome :)

TBH I don't quite understand the crash scenario, would you be willing to include a very simple minimum working example?  The only other classes that could also suffer from this are `ComboBox` and `ColorPicker`, which both inherit from `PopupButton`.  So I think this solution should still fix both.  But since I'm not understanding how to reproduce the issue, I can't say for sure...

crazycarl:
I've never done a Pull Request before ~~noob status here~~ so I'll start by looking up a good tutorial on that.

I'll go hackup one of the examples to demo this crash first.

crazycarl:
I cobbled together this from example2.cpp. Just replace the contents of example2.cpp and you'll see the bug in action. Ohh, and follow the little instructions written on the buttons.

```c
/*
    src/example2.cpp -- C++ version of an example application that shows
    how to use the form helper class. For a Python implementation, see
    '../python/example2.py'.

    NanoGUI was developed by Wenzel Jakob <wenzel.jakob@epfl.ch>.
    The widget drawing code is based on the NanoVG demo application
    by Mikko Mononen.

[...]

PR #327, opened 2018-04-04
Fixed crash caused by PopupButtons when disposing of a window

crazycarl:
Made in response to issue #326


wjakob:
Excellent, thank you!

PR #372, opened 2019-03-06
update to official Eigen 3.3.7, shallow submodule

svenevs:
What are your thoughts on shallow submodules?  The eigen update will reduce many warnings (at least in newer GCC), but the git history on the official mirror is massive.  With a shallow module it clones faster than pybind11 does.

- shallow submodules introduced in 2.10, september 2016
- iirc your benchmark for "oldest support" is debian stretch, [which has 2.11](https://packages.debian.org/search?keywords=git)

Did it on the fork to test: `git clone --recursive https://github.com/svenevs/nanogui.git -b deps/update-eigen`

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -6,7 +6,8 @@
     url = https://github.com/wjakob/glfw
 [submodule "ext/eigen"]
     path = ext/eigen
-    url = https://github.com/libigl/eigen.git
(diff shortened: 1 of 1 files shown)


wjakob:
Super-cool, I'm glad to hear that an official repo is finally available.

PR #383, opened 2019-03-27
std::snprintf is available in MSVC >= 2015

svenevs:
- MSVC 2015 is minimum supported version.
- This is the only CRT secure warning triggered.

This was something that came up in the upcoming cmake stuff.  I'm trying to avoid the need for `_CRT_SECURE_NO_WARNINGS` being part of the public compile definitions, but keeping `NANOGUI_SNPRINTF` around means that we either need to

1. Add `_CRT_SECURE_NO_WARNINGS` to public interface (for eventual `find_package(nanogui)`), or
2. Only add to private compile definitions, but then this forces consumers to additionally disable `_CRT_SECURE_NO_WARNINGS`.

Neither is ideal, (1) may silence warnings that consumers may not actually want to ignore.  But (2) is tedious.

This skirts around the issue quite nicely I think :)


svenevs:
Sigh.  We'll need to keep `_CRT_SECURE_NO_WARNINGS` for things like `fopen` and `sscanf`, but since those happen in source files the new cmake stuff can get away with _private_ compile definitions.

wjakob:
Nice to get rid of the compat header :)

PR #384, opened 2019-03-27
example1: image panel callback bounds need to be checked

### LINKED ISSUE #377: Bug with ImageView widget or just the example1.py, when clicking the thumbnails to pick an image
It might be just an issue with the example1.py, but still worth sharing:

bug

svenevs:
Fixes #377.  Thanks for reporting @gmagno :)

### DIFF
--- a/python/example1.py
+++ b/python/example1.py
@@ -152,8 +152,9 @@ class TestApp(Screen):
         imgView = ImageView(img_window, icons[0][0])
 
         def cb(i):
-            print("Selected item %i" % i)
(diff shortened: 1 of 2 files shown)


wjakob:
Thanks!

svenevs:
Crap.  I introduced a sign compare warning here, just noticed it when working on downstreaming this...sorry.  Do you have a preference for resolve?

- `if (std::size_t(i) < m_images_data.size())`
- `if (static_cast<size_t>(i) < m_images_data.size())`

?

PR #385, opened 2019-03-28
Update license year and fix docs

svenevs:
This should be the last PR before the cmake stuff.  Serves as a test to make sure commits hitting master trigger builds (cmake has associated build docs updates).

RTD now has an [integrations](https://readthedocs.org/dashboard/nanogui/integrations/) tab (news to me).  IIRC the hosting backend service for RTD was changed, so presumably every project needs to re-setup integrations.

- Ref for pygments stuff: [current latest build](https://nanogui.readthedocs.io/en/latest/api/program_listing_file_nanogui_common.h.html), that's not the original sphinx highlighting theme we had, that happens because of changes in sphinx.  This PR sets it back to original (by _not_ setting it in `conf.py`, meaning `sphinx_rtd_theme` gets to set it for us).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,7 +26,7 @@ libnanogui.dylib
 # html documentation specific
 *.pyc
 docs/_build
-docs/doxyoutput
+docs/_doxygen
 docs/api
 
 # entypo update generation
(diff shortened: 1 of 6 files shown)


wjakob:
merged!

svenevs:
Ok this didn't trigger but apparently we needed to reset the default branch to `master`. I just set that and it triggered a build just now so it should work again in the future :)

PR #389, opened 2019-04-03
Fix dangling child widgets from Screen's focus/drag

wardw:
This is a further step along from #341 which itself addresses #332 and the other variants discussed in that PR. 

The essential change from #341 is to let a `Widget` be responsible for it's own removal from `Screen`, as a structured part of its destruction.  Reasoning follows typical RAII semantics:

Stepping back a bit, it strikes me that root of the issue is that `Screen` is holding a reference to `Widget`s for drag/focus, but isn't taking part in their reference counting.  I assume this is because `Screen` really only wants a "weak" reference, but in so far as it's still a reference (and in the absence of implementing alternatives) I suspect it's still probably a good idea that these references are accounted for somewhere..

So - this change is to let a `Widget` itself own this responsibility and make calling `Screen::disposeWidget()` part of a widget's destruction.  To destroy a widget _is_ to "dispose" of it - the two are semantically linked and so this enforces the point structurally.  It enforces the responsibility of a widget to remove itself from the screen.  

-- There is one problem with it as it stands - a `Widget` _is_ a `Screen`, so when the `Screen` is destroyed it will throw.  I've added a temporary workaround to this PR which wraps the call to `Widget::screen()` in a try/catch block: I will leave it up to you whether you might consider `Widget::screen()` returning `nullptr` to signal 'no screen' and make a `Widget` having no screen a valid part of it's semantics.  I've left this PR in draft state - happy to add further if you can't foresee any other problems with this approach.

On the plus side, it does mean we could dispense with the additional `Widget::removeChildHelper()` code and we are back to the single API `Widget::removeChild()` which thanks to @chpatton013 now covers all cases in a consistent way.

Let me know what you think.  There's a test example in [bugfix_test](https://github.com/wardw/nanogui/tree/bugfix_test) (just a variant of example1).

### DIFF
--- a/include/nanogui/screen.h
+++ b/include/nanogui/screen.h
@@ -179,7 +179,7 @@ public:
 
     /* Internal helper functions */
     void updateFocus(Widget *widget);
-    void disposeWindow(Window *window);
+    void disposeWidget(Widget *widget);
     void centerWindow(Window *window);
     void moveWindowToFront(Window *window);
     void drawWidgets();
--- a/include/nanogui/widget.h
+++ b/include/nanogui/widget.h
@@ -274,6 +274,13 @@ protected:
      */
     inline float icon_scale() const { return mTheme->mIconScale * mIconExtraScale; }
 
+private:
+    /**
+     * Convenience function to share logic between both signatures of
+     * ``removeChild``.
+     */
+    void removeChildHelper(const std::vector<Widget *>::iterator& child_it);
+
 protected:
     Widget *mParent;
     ref<Theme> mTheme;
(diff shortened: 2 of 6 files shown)


wardw:
Yes on reflection I think making it part of `removeChild` is better.  You make a good point that should it be possible to move a widget between screens then yes it's true the semantics are separate: add/remove is different from create/destroy.  For me I think that nails it.  It's also less impactful in terms of the scope of the change.

Sorry for the goose chase!  So yes - perhaps better to continue on from #341.

wardw:
Nice library, by the way 👍

svenevs:
> should it be possible to move a widget between screens

For clarity, this was inaccurate (the same Widget instance should not be drawn on two separate screens).  But it is possible to remove it from one screen and add it to another without deleting it :)

> Nice library, by the way +1

Agreed, I am very grateful for NanoGUI (I am not the author) xD

PR #402, opened 2019-05-23
Overload `CreateImage` to support numpy (RGBA uint8) images

martinruenz:
It can be useful to open images from memory in python, for instance to show results generated with numpy.

Such functionality is implemented in this pull request by overloading the function `NVGcontext.CreateImage`

Usage example:
```python3
import PIL
import PIL.Image
import numpy as np
import nanogui

numpy_image = np.random.randint(0,255,[400,400,4]).astype(np.uint8) 
numpy_image[:,:,3] = 1 # Set alpha to 1

nanogui.init()
screen = nanogui.Screen((500, 500), "numpy image example", fullscreen=False)  
window = nanogui.Window(screen, "window") 
window.setLayout(nanogui.GroupLayout())

tex_id = screen.nvgContext().CreateImage(numpy_image, 0)
image_view = nanogui.ImageView(window, tex_id)

screen.setVisible(True)
screen.performLayout()
nanogui.mainloop()
```

### DIFF
--- a/python/nanovg.cpp
+++ b/python/nanovg.cpp
@@ -119,6 +119,16 @@ void register_nanovg(py::module &m) {
         .def("SkewY", &nvgSkewY, "angle"_a)
         .def("Scale", &nvgScale, "x"_a, "y"_a)
         .def("CreateImage", &nvgCreateImage, "filename"_a, "imageFlags"_a)
+        .def("CreateImage",
+             [](NVGcontext *ctx, const py::array_t<uint8_t>& image, int flags){
+                  if(image.ndim() != 3 || image.shape(2) != 4)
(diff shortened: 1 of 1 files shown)


svenevs:
```console
$ resources/check-style.sh
Error: found the following coding style problems:
    python/nanovg.cpp:124:                                                   py::array::forcecast>& image, int flags){
    python/nanovg.cpp:125:                  if(image.ndim() != 3 || image.shape(2) != 4)
The command "resources/check-style.sh" exited with 1.
```

It wants spaces: `){` -> `) {` and `if(` -> `if (`

The example is nice, but do be careful not to create one ever frame ;)

PR #404, opened 2019-05-25
add Theme bindings for python

svenevs:
Partial extract from #317.  Theme bindings easy to add, requested by @jeffsprenger.  Easy for me to rebase modern cmake stuff so figured should just add it now.

Example:

```py
import nanogui

if __name__ == "__main__":
    nanogui.init()
    screen = nanogui.Screen((500, 500), "Theme Bindings :)")
    # NOTE: users should modify the theme before creating additonal
    # widgets.  Most widgets work with setTheme(...), but some do
    # not and will not be fixed in this fork.
    theme = screen.theme()
    # Example: change window focused fill color to red.
    theme.mWindowFillFocused = nanogui.Color(1.0, 0.0, 0.0, 1.0)

    # We changed the screen's theme, any widget added to this as
    # a child inherits a reference to this theme.
    window = nanogui.Window(screen, "Red Window")
    window.setSize((300, 300))

    # Mouse hover the window to see effect of changing mWindowFillFocused.
    screen.performLayout()
    screen.setVisible(True)
    nanogui.mainloop()
```

Refresh: #317 fixed some of the things that happen with dynamic theme changes for specific widgets.  Those will get fixed in mitsuba version off in the future, just need to clarify for Jeff that the most reliable approach to theme customization is to construct with the customized theme rather than call `setTheme` later.


jeffsprenger:
Stephen,

With the python wrap for Theme working, this is how I'm using it within
python.
After creating a Screen object:

        theme = self.theme()

        theme.mStandardFontSize = 20

similar to your proposed approach. Any widgets added to the Screen window
automatically use the same theme.

Jeff.

PR #410, opened 2019-05-31
Run performLayout() again when the combobox gets a new set of items

ermarch:
If the item-list of a combobox changes (multiple times) at run-time it is required to re-calculate the layout again otherwise the popup widget will be stale or empty. Unfortunately there is no way to request performLayout for a single widget so they will all be recalculated.

### DIFF
--- a/src/combobox.cpp
+++ b/src/combobox.cpp
@@ -9,6 +9,7 @@
     BSD-style license that can be found in the LICENSE.txt file.
 */
 
+#include <nanogui/screen.h>
 #include <nanogui/combobox.h>
 #include <nanogui/layout.h>
 #include <nanogui/serializer/core.h>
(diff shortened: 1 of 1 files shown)


biglizards:
I haven't tried it myself, but couldn't you do `somewidget->performLayout(screen->nvgContext())`, since all widgets have a performLayout method?

ermarch:
Good idea, if it works I'll update the pull request.

Erik

ermarch:
Unfortunately this doesn't seem to work as expected.

Mainly because the number of items in the list stays fixed. So updating 
with a different number of items causes a problem.

I'm afraid this might lead to a cascading number of fixes in the 
PerformLayout functions to get it fixed properly.

Erik

PR #418, opened 2019-07-20
Build in stb_image functions as static

ermarch:
to prevent a multiple-defined errors when static linking with nanovg


wjakob:
One small change request: can you just remove "stb_image" from example 4? It is not even used there :)

ermarch:
Sure, I must admit: I've never checked that..
Done.

wjakob:
Merged, thanks!

PR #421, opened 2019-08-30
widgets: add SwitchBox widget

dalerank:
why: switchbox is simple checkbox with drawing switch process


dalerank:
Here it looks now switchbox

